PR #6, opened 2016-02-16
Import initial DB schema from the warehouse

vitorbaptista:
@roll Would you please review this?

The schema is basically the same as you've done at https://github.com/opentrials/warehouse/blob/master/warehouse/database/schema.py, with two small modifications:

* The primary key attribute is called `id`, instead of `uuid`;
* The table names are in plural (i.e. `trials` instead of `trial`).

From what I've seen, these patterns are more common ([rails](http://guides.rubyonrails.org/getting_started.html), [sqlalchemy](http://docs.sqlalchemy.org/en/latest/orm/tutorial.html), [bookshelf.js](http://bookshelfjs.org/)).

The rest of this PR are changed needed to make the tests pass given the new schema.


roll:
@vitorbaptista 
reviewed 
about table naming - it's old holy war - http://stackoverflow.com/questions/338156/table-naming-dilemma-singular-vs-plural-names - I suppose we should use the way better works with ORM and Swagger.

vitorbaptista:
Wow! Didn't know about this eternal war with plurals/singulars. I thought most of frameworks have agreed on plurals (except Django), but probably that's just because of a coincidence: most (all?) frameworks I have experience in use plurals.

It doesn't matter, to be honest. If you prefer, I'd be happy to change back to singulars.

roll:
I think we should go with `plurals` to do not change it without reason and because `bookshelf.js` uses `plurals` in examples - so may be it preferred way to work with this ORM.

vitorbaptista:
Fair enough. Merged :+1:

PR #9, opened 2016-02-24
Add locations, persons and interventions to trials endpoint

vitorbaptista:
Following our [current DB schema](https://github.com/opentrials/api/blob/master/docs/database.md), we have many m-to-n relationships, which require extra tables (which have attributes of their own).

This pull request implements models for 3 of these relationships: locations, persons and interventions. Most of the time was spent working on the structure on how these parts are connected. The main decision I made were:

## How "complete" should the returned Trial be (i.e. if does it contain just its own data, or also data from its relationships)?

Getting data from each relationship has a performance penalty, as it's more data to get from the DB. Although, as we're in the early days, and the data doesn't change much (which makes it easy to cache), I haven't bothered with it: I return everything related to the trial.

The trials are returned in full both in the trials list (`/trials`) and trials details endpoints (`/trials/{id}`). In the future, these endpoint will probably return different values (the list returning less detail).

## How the API response for a Trial will look like with these extra data?

The relationships between (e.g.) Trial and Person have attributes of its own. In the response, I've added:

```js
{
    // ...
    persons: [
        {
            role: 'principal_investigator',  // This comes from the relationship (trials_persons)
            attributes: { /* the attributes */ },
        },
        // ...
    ]
}
```

Every relationship follows this pattern.

## Should we expose every entity?

We have endpoints for trials (`/trials` and `/trials/{id}`). Should we have endpoints for other entities, like persons, as well?

This might make sense down the line, but I would rather not spend time on it before we find a use case where it'll be useful (and do only for the entities that make sense).


vitorbaptista:
Examples of the current API:

http://opentrials-api.herokuapp.com/v1/trials
http://opentrials-api.herokuapp.com/v1/trials/53ff08be-34ff-44e1-9554-c980e6c2dddb

roll:
> Should we have endpoints for other entities, like persons, as well?

I suppose only clients could answer it (like apps or side api users). So if @pwalsh doesn't have it in requirements because of some info he has no reason to add it for now.

PS.
role's `attributes` sounds better than `context` so may be change it also in the schema

pwalsh:
@roll @vitorbaptista there are clear use cases for exposing the relations on their own endpoints (and if there were not, we'd would not need to do the separation in the first place).

Examples:

* Crowdsourced dedupe UI for each relation (primary use case will be built)
* In the main OT app, just browse interventions, just browse conditions

So: yes, I'd go in the opposite direction: expose every entity first.

vitorbaptista:
@pwalsh Agreed. It's mostly a question of approaching it in a top-down or bottom-up way.

About this PR, have you or @roll had the time to take a look into it? Is it good to go?

roll:
@vitorbaptista looks good

vitorbaptista:
Thanks, @roll. Please note that I've removed the unique constraint on `persons.name` column. I have a few issues with it and my tests fixtures and, as two people can have the same name, I ended up removing it.

roll:
@vitorbaptista Yes of course. I've commented it on initial model as temporary or for test purposes.

PR #12, opened 2016-03-07
Search endpoint

vitorbaptista:
This adds the http://opentrials-api.herokuapp.com/v1/search endpoint. This endpoint allows searching on all fields of a trial (including its related entities). For example, http://opentrials-api.herokuapp.com/v1/search?q=united%20states

It also accepts a `per_page` attribute (between 10 and 100), and a `page`.

To create the elasticsearch index, you have to run `npm run reindex`.

The ElasticSearch queries don't do partial matches by default (i.e. `musc` won't match `muscle`), but you can use ES' query language, like `mus*` or `muscle OR heart`, etc.

This PR also removes the `/v1/trials` endpoint. The data it returned can be returned by simply calling `/v1/search` without a query, so it's redundant.


roll:
looks good

vitorbaptista:
Thanks, @roll

PR #15, opened 2016-03-14
Search filters

### LINKED ISSUE #13: Add search filters
Related to https://github.com/opentrials/opentrials/issues/26

vitorbaptista:
_Related to https://github.com/opentrials/opentrials/pull/33. Fixes #13._

There were only two things needed here: add a `/locations` endpoint, so the opentrials.net site could show a list of locations to filter by, and explicitly generating a mapping in elasticsearch.

The mapping was needed because we're using filters as part of the query string. For example, to query the results in a specific location (say Brazil), the user could simply pass `location:Brazil` (http://opentrials-api.herokuapp.com/v1/search?q=location:Brazil). This was inspired by GitHub's search qualifiers (check https://developer.github.com/v3/search/#search-issues), although it doesn't work the same way yet. For example, the default operator is `OR`, which means that a query for `osteoporosis location:Brazil` will returns trials that contain `osteoporosis` or are located in `Brazil`. We can change this afterwards, though.

### DIFF
--- a/api/controllers/locations.js
+++ b/api/controllers/locations.js
@@ -1,0 +1,13 @@
+const Location = require('../models/location');
+
+function list(req, res) {
+  return new Location().fetchAll()
+    .catch((err) => {
+      // FIXME: We're ignoring errors for now, but we should at least log them.
+    })
+    .then(res.json);
+}
+
+module.exports = {
+  list,
+}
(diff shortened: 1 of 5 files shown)


vitorbaptista:
Reviewed by @amercader

PR #16, opened 2016-03-15
Deployment example [NOT FOR MERGE]

roll:
@vitorbaptista some templates for deployment

PR #18, opened 2016-03-18
[WIP] Setup docker

vitorbaptista:
@roll This is for you to review.

This is heavily based on the way you organized #16. There're a few differences, though.

The "deployment" in this PR is pushing to Docker Hub. This was easier than configuring `docker-cloud` in Travis and, as we can configure Docker Cloud to automatically get the new images as they're published to Docker Hub, we keep doing Continuous Deployment. The disadvantage is that changes to the actual `docker-compose.yml` require us to manually re-deploy.

For migrations, we could do as you've done on #16 and run another container that simply runs `npm run migrate && npm run reindex`. I'm not very fond on automatically running migrations, though, as they can be risky. However, considering we're deploying automatically to staging and not production, it's probably OK. We could simply add your idea of a container just to run migrations to the `docker-compose.yml` :+1:

I'm also using a `docker-compose.override.yml`. The idea is that the `docker-compose.yml` would be how the containers should be run in production, while the `docker-compose.override.yml` makes the necessary changes for development. For example, currently there're DB and ElasticSearch contaners in `docker-compose.yml`, but (as you said) we won't be setting up containers for those services. They can be moved to `docker-compose.override.yml`, so they're only used in development.

What do you think?


pwalsh:
@akariv See what @vitorbaptista has here with the override - looks useful for OpenSpending, where we have the same pattern of docker containers for persistent services, for ease of use in local development, and not in  production.

akariv:
We actually have a very similar thing in OpenSpending - we use the `extends` keyword. This allows us to have a more complex inheritance scheme.
In OS, we have a `common.yml` containing all common services and their configurations. We extend this file with `dev.yml` and `production.yml`, both containing specific configurations for production and dev environments - the former containing extra service definition which are only needed in the development environment.
`dev.yml` is also extended with `build.yml`, which contains instructions for also building the docker images.

roll:
I thinks we should use it if it works then iteratively unify deployments between repos and projects (OT, OS). I just didn't have a time to write my thoughts about things like stacks deployment etc.

I suppose the best way will be to start `docker-cloud-example` repo like `oki-py` as RFC (I could do it) and then improve it in discussion with @pwalsh @akariv etc

After we find the best way to use `docker-cloud` we will update our repos accordingly.

vitorbaptista:
> I think we should use it if it works then iteratively unify deployments between repos and projects (OT, OS).

Agreed :+1:

Shall we merge this and see how it works out, then?

roll:
I thinks so also you need `master` to debug it.

PR #25, opened 2016-04-01
Automatic deployment to staging

vitorbaptista:
We're now deploying to Heroku, so there's no need to configure in Travis. I'm also testing the [Heroku Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps) which deploy an instance for each pull request. We might have some problems with the elasticsearch instance, because the plugin environment variable name is `SEARCHBOX_URL`, but we expect `ELASTICSEARCH_URL`. Should be easy to fix, though.


vitorbaptista:
I'm merging this because I need it on master to test the Heroku Review Apps. The Travis failure is because of coveralls (which seems to be happening quite frequently lately :/)

PR #30, opened 2016-04-05
[#65] Return at most 100 pages

vitorbaptista:
There're performance issues on paginating too many results. GitHub limits
1,000 results per query, for example. I would like to do the same, but that
would require validating on two query parameters (`page` and `per_page`). It's
easier to simply limit on a single parameter, so that's what I've done here. If
we have performance issues, we can improve on this solution.

In the worst case, a user would be able to get 100 pages with 100 results each,
totaling 10,000 results.

Fixes opentrials/opentrials#65

Related to https://github.com/opentrials/opentrials/commit/593b0776a2b7b8702cf35d887fd2e703333f279a

### DIFF
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -31,6 +31,7 @@ paths:
           description: The page number
           type: integer
           minimum: 1
+          maximum: 100
           default: 1
         - name: per_page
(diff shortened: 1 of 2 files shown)


pwalsh:
👍

PR #31, opened 2016-04-06
Feature/migration for mapper

roll:
Added 2 migrations:
- added `records.reference` (url normally) and `records.extracted_data` (normalized trial data)
- added meta fields needed by `mapper` to all entity fields


roll:
@vitorbaptista please take a look

migrations/20160406175457_add_meta_fields_to_entities.js
```diff
@@ -0,0 +1,50 @@
...
+      var addTimestamps = knex.schema.table(table, (table) => {
+          table.timestamp('created_at').nullable();
+          table.timestamp('updated_at').nullable();
+          table.specificType('links', 'text[]').nullable();
+          table.specificType('facts', 'text[]').nullable();
+      });
+      var addLinks = knex.schema.raw(
+        'CREATE INDEX '+table+'_links_idx on '+table+' USING GIN(links)')
+      var addFacts = knex.schema.raw(
+        'CREATE INDEX '+table+'_facts_idx on '+table+' USING GIN(facts)')
```
vitorbaptista: Instead of creating the indexes using raw SQL, you could use:

```javascript
table.specificType('links', 'text[]')
  .nullable()
  .index();
```

There's a way to specify the index type to be GIN. It should be easy, although I'm not sure how to use it (would have to test). Please note the structure of the code as well, chaining the methods in newlines...

Also, it won't be useful here, but you can use string interpolation in JavaScript ES6 as `'CREATE INDEX ${table}_links_idx ON ${table} USING GIN(links)'`

migrations/20160406175457_add_meta_fields_to_entities.js
```diff
@@ -0,0 +1,50 @@
...
+    'records',
+    'sources',
+    'trials',
+]
+
+exports.up = (knex) => {
+
+  var operations = []
+  tables.forEach((table) => {
+      var addTimestamps = knex.schema.table(table, (table) => {
```
vitorbaptista: Avoid reusing the same variable name. Maybe renaming the `tables` array to `tableNames` and using `knex.schema.table(tableName, (table) => {`

migrations/20160406175457_add_meta_fields_to_entities.js
```diff
@@ -0,0 +1,50 @@
...
+          table.timestamp('created_at').nullable();
+          table.timestamp('updated_at').nullable();
```
vitorbaptista: Instead of setting those, use `table.timestamps();`. It creates the `created_at` and `updated_at` columns as `nonNullable` and sets them to the current time by default. There might be some problems given that we're not starting from a clean database, but I don't think so.

vitorbaptista:
@roll Apart from the comments in the code itself, avoid using `var` and use either `const` or `let` instead. 

Also, keep in mind that `Promise.all` runs all promises in parallel, so the current code where you create the indexes in a separate promise might break. If you implement my suggestion at https://github.com/opentrials/api/pull/31/files#r58731336, you can keep using `Promise.all`, otherwise you would have to chain the promises manually (similar to https://github.com/opentrials/api/blob/master/tools/create-trials-index.js). I've just opened issue https://github.com/opentrials/opentrials/issues/77 to fix this on the current migration.

roll:
Fixed everything. Thanks. I can't find I way to set default values for postgres `jsonb` and `text[]` though.

roll:
@vitorbaptista 
please re-review - I've changed it based on https://github.com/opentrials/opentrials/issues/78

migrations/20160407111633_create_trialrecords_entity_drop_records.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  const dropTables = knex.schema
+    .dropTableIfExists('trials_records')
+    .dropTableIfExists('records');
+
+  return Promise.all([
+    createTrialrecords,
+    createTrialsTrialrecords,
+    dropTables,
```
vitorbaptista: This needs to be done in sequence, not in parallel. You could probably do something like:

```javascript
dropTables
  .then(createTrialsrecords)
  .then(createTrialsTrialrecords)
```

Maybe you'd need to define an anonymous functions instead (like `then(() => createTrialsrecords)`), but I don't think so.

Also, I think using `trial_records` and `trials_trial_records` would be more consistent, even though I understand that you're probably trying to be consistent with the pattern `FOO_BAR` as relationship between `FOO` and `BAR`.

roll: Yea I think for our complex data model underscores between one entity name is unavailable.

PR #32, opened 2016-04-07
[#66] Set "AND" as the default operator for ElasticSearch

vitorbaptista:
Fixes opentrials/opentrials#66


vitorbaptista:
There's no way to test this on staging, unfortunately, because our seed data is very limited :/

roll:
:+1:

PR #33, opened 2016-04-08
fixed race conditions in migration

roll:
It fixes https://github.com/opentrials/opentrials/issues/77

### DIFF
--- a/migrations/20160216120159_create_initial_schema.js
+++ b/migrations/20160216120159_create_initial_schema.js
@@ -1,5 +1,7 @@
 exports.up = (knex) => {
-  const createSources = knex.schema.createTable('sources', (table) => {
+  const schema = knex.schema;
+
+  schema.createTable('sources', (table) => {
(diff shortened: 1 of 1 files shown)


roll:
@vitorbaptista please take a look

vitorbaptista:
Saving the `knex.schema` instance and returning it is clever, but I feel it makes the code a bit too magical. Maybe a simpler approach would be creating functions like:

```javascript
  const createSourcesTable = (table) => {
    table.uuid('id')
      .primary();
    table.text('name')
      .notNullable();
    table.enu('type', [
      'register',
      'other',
    ]).nullable();
    table.jsonb('data')
      .notNullable();

    table.unique(['name', 'type']);
  };

  knex.source.createTable(createSourcesTable)
    .then(createTrialsTable)
    .then(createFooTable)
    ...
```

However, this solves the issue, so +1 on merging it :+1:

PR #34, opened 2016-04-13
Feature/expose all entities in endpoints

vitorbaptista:


vitorbaptista:
This is ready for review.

migrations/20160413110458_add_gender_and_has_published_results_to_trials.js
```diff
@@ -0,0 +1,21 @@
...
+);
+
+exports.down = (knex) => (
+  knex.schema.table('trials', (table) => {
+    table.dropColumns([
+      'has_published_results',
+      'gender',
+    ]);
+  })
+);
```
roll: We need the same for `trialrecords` table.

roll:
All good except adding a migration for `trialrecords` table.

vitorbaptista:
Good catch! I've added the migration and merged it. Should be in prod in a few minutes.

PR #35, opened 2016-04-15
Remove trials_trialrecords table and add sources and trialrecords seeds

vitorbaptista:
The relationship between Trial and Trialrecords is one-to-many, so there's no
need for a relationship table.

@roll please take a look


roll:
:+1:

PR #36, opened 2016-04-22
Feature/instance functions

borysyuk:


vitorbaptista:
In the async tests (like `test/api/controllers/interventions.js`), you have to either return a promise or accept a `done` callback parameter and call it when you're done. Otherwise, Mocha won't wait for the test result. That's why the current test suite is passing on Travis, but if you see the logs at https://travis-ci.org/opentrials/api/builds/125074792, you'll see a few failures with `Unhandled rejection error`.

ES6 has a quite subtle implicit return value. For example:

```javascript
// this will work fine
it('foo', () => (
  server.inject('/foo')
    .then((response) => response.statusCode.should.equal(200));
));

// this won't
it('bar', () => {
  server.inject('/bar')
    .then((response) => response.statusCode.should.equal(200));
});
```

The `foo` test function is defined with `() => ( /* ... */ )`, while `bar` is defined as `() => { /* ... */ }`. `foo` will implictly return its value, while `bar` won't. You'd have to either change `bar` use the `() => ( /* ... */ )` format or:

```javascript
// this won't
it('bar', () => {
  return server.inject('/bar')
    .then((response) => response.statusCode.should.equal(200));
});
```

This is very subtle and I might end up changing our eslint rules to disallow the `() => ( /* .. */ )` pattern in the future.

This fix will also reveal another problem: the fixtures don't have the `id` defined and aren't saved by default, so you'll have to save them before trying to get them. Don't bother much here, because I've configured another fixture library (actually, factories) in a separate branch (check https://github.com/opentrials/api/blob/feature/expose-trialrecords/test/factory.js). They'll make this easier.

vitorbaptista:
Actually, don't bother with the fix I talked about in https://github.com/opentrials/api/pull/36#issuecomment-214218032. I'll merge https://github.com/opentrials/api/blob/feature/expose-trialrecords/test/factory.js before this code and will fix the tests. It'll be quite easier :+1:

api/controllers/interventions.js
```diff
@@ -0,0 +1,23 @@
+const Intervention = require('../models/intervention');
+
+function getIntervention(req, res) {
+  const id = req.swagger.params.id.value;
+
+  return new Intervention({ id: id }).fetch({ withRelated: Intervention.relatedModels })
```
vitorbaptista: `Intervention.relatedModels` is undefined, so you can change it to just `.fetch()` or `.fetch({})` (if you have problems). The same for all other models.

borysyuk:
@vitorbaptista I have merged our changes and resolved conflicts.

borysyuk:
@vitorbaptista I have added several tests and i got 100% coverage by functions.

vitorbaptista:
Merged in https://github.com/opentrials/api/commit/c052125a128477ca027699ed568fdc0cf6fdf2c8

PR #38, opened 2016-04-25
Feature/updated database docs

roll:


vitorbaptista:
Everything looks fine, I just didn't understand why you linked to a specific migration.

roll:
Kinda versioning - so this docs is actual for this state of `database`.

vitorbaptista:
In that case, I would suggest changing to something like `Last updated on migration 20160415172210`.

roll:
I've already tried  `This document is actual for this migration:` but I like your version better. So updated to `Last updated on migration 20160415172210`

PR #44, opened 2016-05-11
Feature/natural source_id and new problems/interventions fields

roll:
`processors` component requires to have natural id (like `nct` not uuid or autoincr for example) for `sources` table so it will be able to work with priorities on writing (like fda intervention has a more priority than nct interventions etc).

Also it will solve https://github.com/opentrials/opentrials/issues/101 on the next data re-population.

Tested with `explorer` - works OK.


roll:
I will also add to this PR new fields to `problems/interventions` needed by `fda/icd` sources (can't work with migrations in two branches). So renamed and marked as `WIP`.

roll:
@vitorbaptista 
Please take a look.

vitorbaptista:
Could you explain a bit more the reason for this change? From what I understood, this is so in the `processors` you could insert some data using (say) `source_id = 'nct'`, without needing to go to the API database to check that the NCT source has `id = 'ea678007-8749-4d4e-9cbc-087a1e1e95c6'`. Is it correct?

roll:
@vitorbaptista 
Yes, thru all writers passed `source_id` so with natural key priority could be checked without additional queries.

Also it will fix this - https://github.com/opentrials/opentrials/issues/101 

And also my initial idea to have uuid for this entity was just wrong :confused:

PR #45, opened 2016-05-11
Feature/stats endpoint

borysyuk:


test/api/controllers/stats.js
```diff
@@ -0,0 +1,41 @@
...
+  before(clearDB);
+
+  afterEach(clearDB);
+
+  describe('GET /v1/stats', () => {
+
+    it('returns the Stats', () => {
+      let promises = [];
+      promises.push(factory.create('trialWithRelated'));
+      promises.push(factory.create('record'));
```
vitorbaptista: Why not simply:

```javascript
const createData = [
  factory.create('trialWithRelated'),
  factory.create('record'),
]
```

api/controllers/stats.js
```diff
@@ -0,0 +1,95 @@
...
+function getTrialsPerRegistry() {
+  return bookshelf.knex
+    .select(
+      bookshelf.knex.raw('primary_register as registry'),
+      bookshelf.knex.raw('count(primary_register)::int')
+    )
+    .from('trialrecords').groupByRaw('primary_register')
+    .then((rows) => {
+      return rows;
+    });
```
vitorbaptista: As far as I know, this `.then((rows) => { return rows; });` isn't doing anything, so it can be removed safely.

api/controllers/stats.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function getStats(req, res) {
+  let promises = [];
+  promises.push(getTrialsCount());
+  promises.push(getTrialsPerRegistry());
+  promises.push(getTrialsPerYear());
+  promises.push(getTopLocation());
+  promises.push(getDateRegistry());
```
vitorbaptista: Same as https://github.com/opentrials/api/pull/45/files#r62985456

api/controllers/stats.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+const Trials = require('../models/trial');
+const bookshelf = require('../../config').bookshelf;
+const Promise = require('bluebird');
+
+function getTrialsCount() {
+  return Trials.count();
+}
```
vitorbaptista: When you move these methods to the models, this `getTrialsCount` becomes useless: simply use `Trials.count()` directly.

vitorbaptista:
Hey @borysyuk, I've added a few comments on the code. It's looking good, but those methods should be in the model, instead of the controller. The controller shouldn't know the model's structure. It makes it much easier to test each method individually as well.

Please, move those methods in the models that make most sense (`trial`, `record`, ...) and add tests for them.

api/controllers/stats.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const Trials = require('../models/trial');
+const Records = require('../models/record');
+const Locations = require('../models/location');
+const Promise = require('bluebird');
+
+function getStats(req, res) {
+  let promises = [
```
vitorbaptista: This should be `const`. Eslint should have complained about this, not sure why it haven't...

api/controllers/stats.js
```diff
@@ -0,0 +1,31 @@
...
+    Trials.count(),
+    Records.trialsPerRegistry(),
+    Trials.trialsPerYear(),
+    Locations.topLocations(),
+    Records.lastRegistryUpdate(),
+  ];
+
+  Promise.all(promises)
+    .then((results) => {
+    res.json({
```
vitorbaptista: The indent level here is wrong

vitorbaptista:
@borysyuk Thanks for the changes, we're almost there. Just, instead of adding the methods to `module.exports`, add them to the model itself. I should've linked it to you before, sorry, but this is an example of the structure:

* https://github.com/opentrials/opentrials/blob/master/models/user.js
* https://github.com/opentrials/opentrials/blob/master/test/models/user.js

After that, this PR is good to go IMHO :+1:

api/models/record.js
```diff
@@ -58,6 +58,36 @@ const Record = BaseModel.extend({
...
+      bookshelf.knex.raw('count(primary_register)::int')
+    )
+    .from('trialrecords')
+    .groupBy('primary_register')
+    .orderBy('primary_register')
+    .then((rows) => {
+      return rows;
+    });
+  },
+  lastRegistryUpdate: function () {
```
vitorbaptista: From what I understood, `lastRegistryUpdate` is the latest time we've updated data from a registry, right? If that's the case, it feels to me that it should be in the `Source` model, and not here. A sign that it doesn't belong in `Record` is that all attributes are related to the `Source`.

api/models/record.js
```diff
@@ -58,6 +58,36 @@ const Record = BaseModel.extend({
...
+      bookshelf.knex.raw('source_id as id'),
+      bookshelf.knex.raw('sources.name as name'),
+      bookshelf.knex.raw('max(trialrecords.updated_at) as updatedate')
+    )
+    .from('trialrecords')
+    .leftJoin('sources', 'source_id', 'sources.id')
+    .groupBy('source_id', 'sources.name')
+    .then((rows) => {
+      return rows.map((item) => {
+        item.updatedate = (item.updatedate !== null) ? item.updatedate : ''
```
vitorbaptista: How can `updatedate` ever be null? All `trialrecords` should have `created_at` and `updated_at` values. If they don't, there's some error in the data.

Also, the name `updatedate` is a bit too generic. It sounds like the same as `updated_at`, but it isn't. Maybe change it to `latest_update_date` or something similar? Make sure you keep the pattern of using underscores in the column names.

api/models/trial.js
```diff
@@ -97,6 +97,19 @@ const Trial = BaseModel.extend({
...
+  return bookshelf.knex
+    .select(
+      bookshelf.knex.raw('to_char(registration_date, \'YYYY\')::int as year'),
+      bookshelf.knex.raw('count(registration_date)::int')
+    )
+    .from('trials')
+    .groupByRaw('year')
+    .orderBy('year')
+    .then((rows) => {
+      return rows;
```
vitorbaptista: This last `.then()` isn't doing anything, so should be safe to remove

test/factory.js
```diff
@@ -128,6 +128,10 @@ factory.define('record', Record, Object.assign({}, trialAttributes, {
   source_url: factory.sequence((n) => `http://source.com/trial/${n}`),
   source_data: JSON.stringify({}),
 }), {
+  afterBuild: (record, attrs, callback) => {
+    record.hasTimestamps = false;
+    callback(null, record);
+  },
```
vitorbaptista: Why are you changing it here? If it's a way to make sure we keep consistent timestamps across tests, you could use `sinon` to mock the timers. Take a look at https://github.com/opentrials/opentrials/blob/feature/contribute-data/test/agents/s3.js#L12-L22 for an example.

test/.eslintrc
```diff
@@ -1,3 +1,7 @@
 ---
 env:
   mocha: true
+rules:
+  no-param-reassign:
+    - 2
+    - props: false
```
vitorbaptista: I think this change was to allow you to change the `record.hasTimestamps` later. If so, please check https://github.com/opentrials/api/pull/45/files#r63379124. Maybe this isn't needed after all.

It's OK to change the eslint rules when they don't make sense, but try to avoid changing globally. You can ignore rules for specific lines or files. For example, check https://github.com/opentrials/opentrials/blob/feature/contribute-data/models/base.js#L13

test/api/controllers/stats.js
```diff
@@ -0,0 +1,31 @@
...
+  describe('GET /v1/stats', () => {
+
+    it('returns the Stats', () => {
+      let promises = [
+        factory.create('trialWithRelated'),
+        factory.create('record'),
+      ];
+      Promise.all(promises).then((result) => {
+        return server.inject('/v1/stats')
+          .then((response) => {
```
vitorbaptista: as you're not using the result from the factories directly, re-arrange this to:

```javascript
return factory.create('trialWithRelated')
  .then(() => factory.create('record'))
  .then(() => server.inject('/v1/stats'))
  .then((response) => {
    // ...
  })
```

The important thing here is to remember to return the promise, otherwise the tests will become flaky because Mocha don't have access to the Promise you're creating, so it can't wait for it to be done. Sometimes it'll be done before Mocha moves on to another test, sometimes it won't.

Also, avoid nested promises when you can. They make the code harder to follow.

test/api/models/location.js
```diff
@@ -23,4 +23,20 @@ describe('Location', () => {
         })
     })
   });
+
+  describe('topLocations', () => {
+    it('is an empty array if there\'re none', () => {
+      new Location().topLocations().then((result) => {
```
vitorbaptista: Return the promise

test/api/models/record.js
```diff
@@ -42,4 +42,51 @@ describe('Record', () => {
       });
     });
   });
+
+  describe('trialsPerRegistry', () => {
+    it('is an empty array if there\'re none', () => {
+      new Record().trialsPerRegistry().then((result) => {
```
vitorbaptista: Forgot to return

test/api/models/record.js
```diff
@@ -42,4 +42,51 @@ describe('Record', () => {
...
+              { registry: 'primary_register1', count: 1 },
+            ]);
+          });
+        });
+    });
+  });
+
+  describe('lastRegistryUpdate', () => {
+    it('is an empty array if there\'re none', () => {
+      new Record().lastRegistryUpdate().then((result) => {
```
vitorbaptista: Forgot to return

test/api/models/trial.js
```diff
@@ -219,4 +219,29 @@ describe('Trial', () => {
         });
     });
   });
+
+  describe('trialsPerYear', () => {
+    it('is an empty array if there\'re none', () => {
+      new Trial().trialsPerYear().then((result) => {
```
vitorbaptista: Forgot to return

test/api/models/location.js
```diff
@@ -23,4 +23,20 @@ describe('Location', () => {
...
+        should(result).deepEqual([]);
+      });
+    });
+
+    it('returns 10 top locations', () => {
+      return factory.createMany('trialWithRelated', 20).then((trials) => {
+        new Location().topLocations().then((result) => {
+          should(result.length).equal(10);
```
vitorbaptista: Assert the actual return value

api/models/record.js
```diff
@@ -58,6 +58,36 @@ const Record = BaseModel.extend({
...
+  lastRegistryUpdate: function () {
+  return bookshelf.knex
+    .select(
+      bookshelf.knex.raw('source_id as id'),
+      bookshelf.knex.raw('sources.name as name'),
+      bookshelf.knex.raw('max(trialrecords.updated_at) as updatedate')
+    )
+    .from('trialrecords')
+    .leftJoin('sources', 'source_id', 'sources.id')
+    .groupBy('source_id', 'sources.name')
```
vitorbaptista: Why grouping on both ID and Name? ID should be enough...

api/models/record.js
```diff
@@ -58,6 +58,36 @@ const Record = BaseModel.extend({
...
+  trialsPerRegistry: function () {
+  return bookshelf.knex
+    .select(
+      bookshelf.knex.raw('primary_register as registry'),
+      bookshelf.knex.raw('count(primary_register)::int')
+    )
+    .from('trialrecords')
+    .groupBy('primary_register')
+    .orderBy('primary_register')
+    .then((rows) => {
```
vitorbaptista: Can safely remove this `.then()`

api/models/location.js
```diff
@@ -16,6 +16,19 @@ const Location = BaseModel.extend({
...
+  topLocations: function () {
+    return bookshelf.knex
+      .select(
+        bookshelf.knex.raw('location_id as id'),
+        bookshelf.knex.raw('locations.name as name'),
+        bookshelf.knex.raw('count(location_id)::int')
+      )
+      .from('trials_locations')
+      .leftJoin('locations', 'location_id', 'locations.id')
+      .groupBy('location_id', 'locations.name')
```
vitorbaptista: Why group by both id and name? ID should be enough.

vitorbaptista:
Hi @borysyuk, I've added a few more comments.

Apart from those, please make sure you're following the coding style of the rest of the code. For example, the definition of the `sourceRelatedToSeveralRecords` factory where you put each parameter in a separate line. There're also a few places where the indentation is wrong, which might be caused by your IDE.

api/controllers/stats.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const Trials = require('../models/trial');
+const Records = require('../models/record');
+const Locations = require('../models/location');
+const Promise = require('bluebird');
+
+function getStats(req, res) {
```
vitorbaptista: Make sure you also catch any errors. If you don't do so, HapiJS will just hang waiting for a reply (which will eventually timeout). Check the other controllers to see how you could do it.

vitorbaptista:
Ah! This is a tiny thing, but overall I have avoided explictly requiring `bluebird` and using its promises, unless there's something specific that I need (e.g. `.promisify()`). In Node, you can simply use `Promise` as is, without a `const Promise = require('bluebird')` call before.

api/models/record.js
borysyuk: when i put ID only i got error :
`Unhandled rejection error: column "sources.name" must appear in the GROUP BY clause or be used in an aggregate function`

api/swagger/swagger.yaml
```diff
@@ -248,7 +248,93 @@ paths:
...
+          $ref: '#/definitions/TrialsPerRegistry'
+      trialsPerYear:
+        type: array
+        items:
+          $ref: '#/definitions/TrialsPerYear'
+      topLocations:
+        type: array
+        items:
+          $ref: '#/definitions/TopLocations'
+      dateRegistry:
```
vitorbaptista: Please rename this to `sourcesLatestUpdatedDate`. Rename also `latest_update_date` to `latest_updated_date`. Sorry I'm being picky here, but as these attributes are part of our public API, it's difficult to change names once we pick one.

vitorbaptista:
Merged in https://github.com/opentrials/api/commit/7b864daaa4df560886a996aac302a8f18943d3a4 and https://github.com/opentrials/api/commit/382f6d9fec5e44a95c31d502a454e78925f87f3e

Thanks, @borysyuk!

PR #48, opened 2016-05-18
[#30] Rename problems to conditions

vitorbaptista:
opentrials/opentrials#30


api/models/condition.js
```diff
@@ -0,0 +1,21 @@
...
+
+const bookshelf = require('../../config').bookshelf;
+const BaseModel = require('./base');
+
+const Condition = BaseModel.extend({
+  tableName: 'conditions',
+  visible: [
+    'id',
+    'name',
+    'type',
```
vitorbaptista: `type` now is just `condition` or `other`, as far as I know. Maybe we should also remove this attribute? /cc @roll

roll: of course remove :+1:

roll:
:+1:

api/models/condition.js
vitorbaptista: Done :+1:

PR #50, opened 2016-05-19
Feature/additional fields in trial model

borysyuk:


borysyuk:
https://github.com/opentrials/opentrials/issues/152

roll:
May be I didn't get it right but:
```
_convertIdentifiers: function(attributes) {
     return [ attributes.primary_id ];
 },
```
doesn't look like a function returning all ids like https://github.com/opentrials/opentrials/issues/152#issuecomment-220243443. 
Sorry if I'm wrong. 

If it's so problematic on API side may really you will just add a migration `identifiers jsonb -> array` and I'll fill it as array?

vitorbaptista:
> If it's so problematic on API side may really you will just add a migration identifiers jsonb -> array and I'll fill it as array?

@roll Yeah, I think that we're spending too much time on a suboptimal solution. Maybe we should simply convert `identifiers` to an array in the database and let the processors populate them.

@borysyuk Sorry for the change after you've started working. Could you remove the logic of altering the identifiers and:
1. Add a `alter_identifiers_type_to_array` migration that changes `trial.identifiers` and `record.identifiers` to an array;
2. Add sample identifiers to the `seeds/trial.js` (so we can try stuff);
3. Expose the `identifiers` in the JSON response.

You might face some issues with BookshelfJS/KnexJS support for arrays. I'm not sure how well they're supported. If it's too much hassle, feel free to keep the column as a jsonb, but then consider that its contents will be an array of strings.

vitorbaptista:
Merged as https://github.com/opentrials/api/commit/80630ee6619dfa7f4e69d869aa05317bace8f139. Thanks!

PR #53, opened 2016-05-26
Add urls to entities

vitorbaptista:
Every entity now has an `url` attribute which points to their API endpoint. This is needed for https://github.com/opentrials/opentrials/issues/124.

@roll This is blocking @smth, so I'm merging it. Please take a look when you can :+1:


roll:
:+1:

PR #54, opened 2016-05-27
fixed records unique constraint

roll:
Current constraint is wrong because for example `source_id=ictrp` having records from other primary registries (so database re-population was in write errors).


roll:
@vitorbaptista 
please review (feel free to merge and push to production without me=)

vitorbaptista:
@roll merged, will deploy later today

PR #55, opened 2016-05-27
Add "/trials/{id}/records" endpoint and "has_discrepancies" attribute in Trial

vitorbaptista:
@roll This is a proposal on how to display the data discrepancies. Here I've added a `/trials/{id}/records` endpoint which returns a list of the Trial's records, and added a `has_discrepancies` attribute to the `Trial` record.

The `has_discrepancies` attribute is to allow searching for trials with or without discrepancies (and change the UI accordingly). The records list is used in the Explorer to build the discrepancies page (~~I'll send a PR in a few minutes~~ PR in https://github.com/opentrials/opentrials/pull/184).

What do you think? The reason I started writting it was that doing this in the API/Explorer levels allow us to change more easily how we display/calculate discrepancies than in the lower levels (processor). The `has_discrepancies` attribute might have a low performance, but as our data mostly doesn't change, we can solve it by caching later.


roll:
I think good solution for now :+1:

PR #56, opened 2016-05-30
[#98] Add constraint for trials and records .recruitment_status as enum

vitorbaptista:
This makes sure we only add valid values to the DB.

opentrials/opentrials#98


roll:
:+1:

PR #57, opened 2016-05-30
Fix bug where the API spec required some fields that could be null

vitorbaptista:
This was introduced by a58871267e3db61df8ea6e1dd83d1bc688325fbc


roll:
:+1:

PR #60, opened 2016-05-31
[#192] Ignore Trials without registration_date on yearly trial count

vitorbaptista:
Fixes opentrials/opentrials#192


roll:
:+1:

PR #61, opened 2016-06-15
[#194] Add trial.discrepancies and remove trial.has_discrepancies

vitorbaptista:
If we don't calculate the discrepancies in the API, we'll have to do it in
the Explorer, which causes duplication because the code to check if a trial has
discrepancies is quite similar to the one to generate the discrepancies list.

opentrials/opentrials#194


roll:
:+1:

PR #62, opened 2016-06-17
[#206] Ignore whitespaces and punctuation when calculating discrepancies

vitorbaptista:
Fixes opentrials/opentrials#206


roll:
:+1:

PR #63, opened 2016-06-22
Feature/make source_url, abstract nullable

roll:
- to fix https://github.com/opentrials/opentrials/issues/204
- to have a possibility to write HRA data as publications

### DIFF
--- a/migrations/20160622144924_make_source_url_nullable.js
+++ b/migrations/20160622144924_make_source_url_nullable.js
@@ -1,0 +1,34 @@
+'use strict';
+
+const tableNames = [
+  'records',
+  'publications',
+];
+
+const fieldNames = [
+  'source_url',
+];
+
+exports.up = (knex) => {
+  const schema = knex.schema;
+
+  tableNames.forEach((tableName) => {
+    fieldNames.forEach((fieldName) => {
+      schema.raw(`ALTER TABLE ${tableName} ALTER COLUMN ${fieldName} DROP NOT NULL`);
+    });
+  });
+
+  return schema;
+};
+
+exports.down = (knex) => {
+  const schema = knex.schema;
+
+  tableNames.forEach((tableName) => {
+    fieldNames.forEach((fieldName) => {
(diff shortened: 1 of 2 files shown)


roll:
@vitorbaptista 
please take a look

Will it be OK for frontend to have `null` in `source_url` for some rows?

vitorbaptista:
@roll I thought the research summaries would be added to `publications.abstract`? If so, I don't understand why changing it to `NOT NULL`. Also, I'm a bit concerned with having records without a `source_url`. As far as I know, all data that we're gathering are available online, right? Even though it might be behind a private API (e.g. HRA), they do provide a public-facing URL. Is it difficult (or impossible) to get those?

It seem to me that what we need is creating a new column (like `human_source_url` or similar) pointing to the HTML source. This way we keep the machine-readable URL as is while also having the human-readable link.

roll:
I'm not sure it's possible to link any data we have and will be have - for example fda stuff from some zip archives etc. So have it nullable seems to be correct.

pwalsh:
Btw I see no problem with a link to an API endpoint or url in general that is not publicly accessible. It is still a link to the source.

roll:
@pwalsh 
Even for example for HRA?

vitorbaptista:
@roll Good point, we might end up with data where we can't link to a source. I don't think we'll have this issue with FDA, though, as all PDFs should be accessible online, it just might be harder to generate the URL. Even for data contributions, we might link to the researcher's page or something similar. However, please correct me if I'm wrong, but everything we have now has a public-accessible URL.

About linking to an API/inaccessible-URLs, I think we should keep that information in our database, but it shouldn't be the default place a person is sent to when they're looking for our sources. Considering our target users, I'd guess linking to a XML or JSON will be quite confusing.

roll:
Let's:
- close it for now
- then add something like `source_url_type`
- will be showing source url as we will decide later using this type

roll:
https://github.com/opentrials/opentrials/issues/238

PR #64, opened 2016-07-05
Add API docs in /v1/docs endpoint

vitorbaptista:
This is generated using Swagger UI. Unfortunately, there's no way to configure
the swagger file's path in Swagger UI, so we need to directly change it in the
code. To avoid doing this change manually, I've added a small bit of code that
does it when starting the server. It's ugly and hacky, but it works for now.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -10,6 +10,7 @@ if (!process.env.ELASTICSEARCH_URL) {
 const elasticsearch = require('elasticsearch');
 const path = require('path');
 const good = require('good');
+const inert = require('inert');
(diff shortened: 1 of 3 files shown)


vitorbaptista:
You can check this by going at https://ot-api-next-pr-64.herokuapp.com/v1/docs/. The default isn't working because it's trying to get the `swagger.yaml` file from a HTTP server (which is blocked at least in my browser). However, you can change the value to https://ot-api-next-pr-64.herokuapp.com/v1/swagger.yaml and everything should work fine.

roll:
It works :+1:

PR #66, opened 2016-07-13
[#254] Returns documents and change relationship between it and trials

vitorbaptista:
We had a `trials_documents` table to generate a many-to-many relationship
between trials and documents. However, a single document can't cite multiple
trials (AFAIK), so we can simplify that to a 1-to-many relationship. This means
we can remove the `trials_documents` table.

I also removed the `documents.slug` column, as we're not using it, and added
the `type` (e.g. csr, epar_segment, ...) and `url` columns. For now, the trials
endpoint returns an array of these documents with their `name`, `type` and
`url`.

opentrials/opentrials#254

### DIFF
--- a/api/models/document.js
+++ b/api/models/document.js
@@ -1,0 +1,15 @@
+'use strict';
+
+const bookshelf = require('../../config').bookshelf;
+const BaseModel = require('./base');
+
+const Document = BaseModel.extend({
+  tableName: 'documents',
+  visible: [
+    'type',
+    'name',
+    'url',
+  ],
+});
+
+module.exports = bookshelf.model('Document', Document);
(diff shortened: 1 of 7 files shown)


nightsh:
OK @vitorbaptista, took me a while to get to review and test both the API PR and the Explorer, all seems fine. If you don't have any further comments, I would merge them both.

vitorbaptista:
Thanks, @nightsh! @roll Could you check this as well? It changes the `documents` table (which would require changes in the processors)

roll:
@vitorbaptista 
Thanks for checking about processors. We don't have one using `documents` so far. 

Also I've added a few comments to the corresponding issue.

LGTM beside those comments.

nightsh:
Turns out somwhere.com is a real website. Maybe go for example.org instead?

vitorbaptista:
@nightsh Nice, I didn't know example.com and example.org are reserved domains to be used in these cases (http://www.iana.org/domains/reserved). I'll start using them now.

PR #67, opened 2016-08-09
[#301] Don't index entities not linked to trials

nightsh:
After several iterations, I found this current implementation as being the fastest and more reliable of all the different ideas I had for it. No need for dedicated methods, helpers etc., as all our autocomplete models now have a relationship with trials.

~~Changed the console text a bit too, in order to reflect the total number that is just being processed for indexing, and not necessarily everything is going to be (re)indexed.~~ I tried to play with the console text a bit, but I managed to get eslint angry instead. :fearful:

### DIFF
--- a/tools/create-trials-index.js
+++ b/tools/create-trials-index.js
@@ -299,6 +299,9 @@ function bulkIndexEntities(entities, index, indexType) {
   }
 
   const bulkBody = entities.models.reduce((result, entity) => {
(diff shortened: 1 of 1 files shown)


vitorbaptista:
@nightsh Missed this PR. Is it ready for review?

vitorbaptista:
Great job! The solution was simpler than I expected :+1:

PR #68, opened 2016-08-18
Changes to the database schema needed for the FDA processor

vitorbaptista:
This adds the new tables and changes necessary for the FDA processor. Here I'm just creating/changing the tables, and not adding new entities and API endpoints yet. Related to https://github.com/opentrials/collectors/pull/34 and https://github.com/opentrials/processors/pull/38.

opentrials/opentrials#302

### DIFF
--- a/migrations/20160818182848_create_fda_approvals_and_add_relationship_with_documents.js
+++ b/migrations/20160818182848_create_fda_approvals_and_add_relationship_with_documents.js
@@ -1,0 +1,44 @@
+'use strict';
+
+exports.up = (knex) => (
+    knex.schema
+      .createTable('fda_approvals', (table) => {
+        table.text('id').primary();
+
+        table.uuid('intervention_id')
+          .notNullable()
+          .references('interventions.id');
+        table.integer('supplement_number')
+          .notNullable();
+        table.text('type')
+          .notNullable()
+          .index();
+        table.date('action_date')
+          .notNullable();
+        table.text('notes')
+          .nullable();
+        table.timestamps();
+
+        table.unique(['intervention_id', 'supplement_number']);
+      })
+      .table('documents', (table) => {
+        table.text('fda_approval_id')
+          .nullable()
+          .references('fda_approvals.id');
+      })
+      .raw('ALTER TABLE documents ALTER COLUMN trial_id DROP NOT NULL')
+      .raw(`ALTER TABLE documents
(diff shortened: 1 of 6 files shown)


migrations/20160816113418_add_application_number_to_interventions.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+exports.up = (knex) => (
+    knex.schema.table('interventions', (table) => {
+      table.text('application_number')
```
roll: `fda_application_number`?

migrations/20160818182848_create_fda_approvals_and_add_relationship_with_documents.js
```diff
@@ -0,0 +1,44 @@
...
+        table.unique(['intervention_id', 'supplement_number']);
+      })
+      .table('documents', (table) => {
+        table.text('fda_approval_id')
+          .nullable()
+          .references('fda_approvals.id');
+      })
+      .raw('ALTER TABLE documents ALTER COLUMN trial_id DROP NOT NULL')
+      .raw(`ALTER TABLE documents
+            ADD CONSTRAINT trial_id_xor_fda_approval_id_check CHECK (
```
roll: Other options is to have `entity_type` (`trial/fda_approval`) and `entity_id` instead of this a little bit weird situation with `trial_id/fda_approval_id`. Later documents could be linked to `conditions`, `organisations` etc.

migrations/20160816113418_add_application_number_to_interventions.js
vitorbaptista: Good point. Changed :+1:

migrations/20160818182848_create_fda_approvals_and_add_relationship_with_documents.js
vitorbaptista: Using a polymorphic association was my first thought, but I decided against it because it opens another can of worms. For example, with it we lose the database's foreign check constraints, so there could be `documents` that with an inexistent `intervention_id` (or `trial_id`).

I agree that adding the `intervention_id` column to `documents` is a bit odd, and is a pattern that obviously won't scale if `documents` start to need relationships to other entities as well. However, as this is the simplest change to achieve what we need now, and it's possible to refactor the relationship into a different pattern whenever we need to expand it, I think this solution is a good compromise for now.

roll:
:+1:

PR #70, opened 2016-08-23
[#278] Ignore EUCTR records when calculating discrepancies

vitorbaptista:
They cause too much trouble because they have translations in different
languages, the `registration_date` is different for each country, the
`target_sample_size` is specific to each country, etc.. To avoid those, we'll
just ignore EUCTR completely for now.

Fixes opentrials/opentrials#278


nightsh:
Looking good to me 👍

PR #71, opened 2016-08-25
[#267] Add "status" and change "recruitment_status" enum in trials and records

vitorbaptista:
Now both a Trial and a Record have "status" that can be "ongoing", "withdrawn",
"suspended", "terminated", "complete" and "other", while "recruitment_status"
can be "recruiting", "not_recruiting", "unknown" and "other". This allows us to
know both the overall status of the trial and its recruitment status.

opentrials/opentrials#267


roll:
👍

PR #73, opened 2016-08-30
Add ON DELETE CASCADE to all trials relationship tables

vitorbaptista:
Added this only for the "trial_id", and not the other foreign keys (e.g.
publication_id on trials_publications) to be safer. Also, removing a trial with
all those different relations is cumbersome, but removing a (say) publication
isn't because we just need to remove the trials_publications referencing it.


roll:
:+1:

PR #74, opened 2016-08-30
[#354] Update fields used to calculate discrepancies

vitorbaptista:
@nightsh I'm merging this before your review because I'd like to calculate the new discrepancies today. Please do review when you see this and we can change as needed.

### DIFF
--- a/api/models/record.js
+++ b/api/models/record.js
@@ -26,6 +26,7 @@ const Record = BaseModel.extend({
     'status',
     'recruitment_status',
     'registration_date',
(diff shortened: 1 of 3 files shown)


nightsh:
I reviewed the code, but haven't followed all the discussion leading to the implementation. I understand the decisions behind it though and I have no objections for the diffs here. I'm happy for not blocking you, especially if I would've given a 👍 anyway ;)

nightsh:
OK, I've read the discussion too. Another 👍

PR #75, opened 2016-08-31
[#361] Add fda_applications and its references

vitorbaptista:
We now have a new "fda_applications" table which will contain metadata for
every FDA application. The "interventions" and "fda_approvals" tables now
references it through their "fda_application_id" columns.

The complexity of this migration is to maintain the existing relationships.
For this reason, I had to set all "fda_applications" columns as nullable, even
though they shouldn't be null. This way, I can create the rows just needing the
IDs (i.e. the FDA application numbers), which we already have in the DB.

When this migration is applied and we have all data in the DB, we should add
the "NOT NULL" constraints.

opentrials/opentrials#361

### DIFF
--- a/migrations/20160831135725_create_fda_applications.js
+++ b/migrations/20160831135725_create_fda_applications.js
@@ -1,0 +1,55 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema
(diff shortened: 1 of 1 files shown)


roll:
:+1:

PR #77, opened 2016-09-01
[#364] Reduce memory usage of the reindexing process

vitorbaptista:
I did so by offloading the removal of entities without trials to the database,
instead of doing it with NodeJS. With these changes, there's no need to query
for the actual trials as well, so it reduces the number of SQL queries in half.

I haven't tested the difference in memory usage, but considering we're both not
loading the trials in memory and loading just the entities we need, I'd imagine
it to be quite significant. Hopefully enough to avoid the Out of Memory errors
we're seeing.

Fixes opentrials/opentrials#364

### DIFF
--- a/tools/create-trials-index.js
+++ b/tools/create-trials-index.js
@@ -321,9 +321,9 @@ function bulkIndexEntities(entities, index, indexType) {
 
 function indexModel(model, index, indexType, _queryParams, fetchOptions) {
   return model.count().then((modelCount) => {
-    const bufferLength = 1000;
+    const batchSize = 1000;
     console.info(
-      `${modelCount} entities being indexed in "${index}/${indexType}" (${bufferLength} at a time).`
+      `${modelCount} entities being indexed in "${index}/${indexType}" (${batchSize} at a time).`
     );
     let offset = 0;
     let chain = Promise.resolve();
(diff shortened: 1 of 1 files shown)


tools/create-trials-index.js
```diff
@@ -358,13 +359,25 @@ function indexAutocompleteModel(model, indexType) {
     model,
     'autocomplete',
     indexType,
-    { columns: ['id', 'name'], withRelated: 'trials' }
+    {
+      // Filter out entities without trials
+      innerJoin: [
```
nightsh: Would be interesting to see if this approach, while consuming more resources for the join, will still be under the footprint Node currently has. Looking forward! :)

tools/create-trials-index.js
```diff
@@ -322,7 +319,7 @@ function bulkIndexEntities(entities, index, indexType) {
   return result;
 }
 
-function indexModel(model, index, indexType, fetchOptions) {
+function indexModel(model, index, indexType, _queryParams, fetchOptions) {
   return model.count().then((modelCount) => {
     const bufferLength = 1000;
```
nightsh: Any reason we can't name this thing here `const limit`?

nightsh:
I'll give it a 👍 , but can you take a look at the line comments before merging? No blockers, just small observations.

tools/create-trials-index.js
vitorbaptista: I think `bufferLength` is clearer than `limit` because it's more specific. A variable `limit` can be many things (e.g. maybe we're just indexing `limit` entities?), but `bufferLength` tells you it's actually a buffer. Reading it again now, should probably have named it to `batchSize` or similar. Will do this :+1:

PR #79, opened 2016-09-06
[#382] Remove `url` from Locations API response

vitorbaptista:
Locations have no `/v1/locations/<id>` route, so the URLs don't exist.

Fixes opentrials/opentrials#382


vitorbaptista:
@nightsh Merging this as well, as we're starting to have too many open PRs. Please take a look when you can.

nightsh:
I think you missed a `url` on autocomplete, I'll remove it in my Swagger PR ;)

vitorbaptista:
@nightsh Good catch :+1:

PR #80, opened 2016-09-06
[#287] Swagger improvements

nightsh:
Changed:

* `operationId` was defined multiple times as `get`, which is invalid according to swagger standards. This was changed in both the API and client and both PRs should be merged at once. In the client, the `agent/*` code was modified to reflect the API changes.

Added:

* More documentation in the `description` tags of the entities

### DIFF
--- a/api/controllers/conditions.js
+++ b/api/controllers/conditions.js
@@ -21,5 +21,5 @@ function getCondition(req, res) {
 }
 
 module.exports = {
-  get: getCondition,
+  getCondition,
 }
--- a/api/controllers/interventions.js
+++ b/api/controllers/interventions.js
@@ -21,5 +21,5 @@ function getIntervention(req, res) {
 }
 
 module.exports = {
-  get: getIntervention,
+  getIntervention,
 }
--- a/api/controllers/organisations.js
+++ b/api/controllers/organisations.js
@@ -21,5 +21,5 @@ function getOrganisation(req, res) {
 }
 
 module.exports = {
-  get: getOrganisation,
+  getOrganisation,
 }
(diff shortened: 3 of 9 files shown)


vitorbaptista:
@nightsh I researched a bit the issue with multiple `operationId` with same name and I found that:

* Our Swagger file is valid according to the official swagger validator: http://online.swagger.io/validator?url=https://raw.githubusercontent.com/opentrials/api/master/api/swagger/swagger.yaml validate swagger
* There's a tutorial by Apigee, which are one of the main contributors to Swagger, that ignores those errors as well (but for a different reason): https://community.apigee.com/articles/25237/heres-why-and-how-you-can-automate-generation-of-a.html

However, the spec itself describes `operationId` as:

> Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.
>
> http://swagger.io/specification/

Which is quite odd if you ask me. However, I agree with you: we should follow the spec.

nightsh:
I have only read the spec, so now I find the whole matter a bit weird. I think following the spec should be the safe zone, but we probably should find a way to keep an eye on it for related changes. I expect in the future either their validator or spec would change.

So, question is now: what should we do with this diff, and the explorer one? I can sync them with master and prepare them for merging, if you agree, or if you have any changes in mind... I'd like to see them gone from the open PRs list :)

vitorbaptista:
@nightsh This PR ended up getting out of sync with master, so we'll need to rebase it, fix the conflicts and merge

nightsh:
This is tied to the [corresponding explorer PR](https://github.com/opentrials/opentrials/pull/384).

vitorbaptista:
Merged as 096dccfbbf8c0d8baee16c9c75ee511061536c30

PR #81, opened 2016-09-07
[#281] Add "url" and "terms_and_conditions_url" to Source

vitorbaptista:
opentrials/opentrials#281

### DIFF
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -294,6 +294,23 @@ paths:
           schema:
             $ref: "#/definitions/ErrorResponse"
 
+  /sources:
+    x-swagger-router-controller: sources
+    get:
+      tags:
+        - sources
+      description: Returns list of sources
+      operationId: list
+      responses:
+        "200":
+          description: Success
+          schema:
+            $ref: "#/definitions/Source"
+        default:
+          description: Error
+          schema:
+            $ref: "#/definitions/ErrorResponse"
+
   /stats:
     x-swagger-router-controller: stats
     get:
@@ -809,6 +826,11 @@ definitions:
       id:
         type: string
       name:
+        type: string
+      url:
+        type: string
+        description: URL to source's website
+      terms_and_conditions_url:
         type: string
       type:
         type: string
--- a/api/controllers/sources.js
+++ b/api/controllers/sources.js
@@ -1,0 +1,18 @@
+'use strict';
+
+const Source = require('../models/source');
+
+function listSources(req, res) {
+  return new Source().fetchAll()
+    .catch((err) => {
+      res.finish();
+      throw err;
+    })
+    .then((sources) => {
+      res.json(sources);
+    });
+}
+
+module.exports = {
+  list: listSources,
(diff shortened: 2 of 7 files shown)


api/swagger/swagger.yaml
```diff
@@ -294,6 +294,23 @@ paths:
           schema:
             $ref: "#/definitions/ErrorResponse"
 
+  /sources:
+    x-swagger-router-controller: sources
+    get:
+      tags:
+        - sources
+      description: Returns list of sources
+      operationId: list
```
nightsh: I'd go with `getSources` here, to comply with the latest Swagger updates (I really need to remove that WIP tag from the PR). This should be reflected in the explorer agent code.

api/swagger/swagger.yaml
```diff
@@ -810,6 +827,11 @@ definitions:
         type: string
       name:
         type: string
+      url:
+        type: string
+        description: URL to source's website
+      terms_and_conditions_url:
```
nightsh: No description?

nightsh:
For more swagger samples related to the issues I've been pointing out in the line comments, see the diff on this PR https://github.com/opentrials/api/pull/80/files

api/swagger/swagger.yaml
vitorbaptista: I was going to add a comment on this change. I like using general names like `list` or `sources` because when you're using the `swagger.yaml` in a client you can do `client.sources.list()` instead of `client.sources.getSources()`. This makes it more succinct , as in `client.persons.get(1)` and `client.organisations.get(1)` instead of `client.persons.getPerson(1)` and `client.organisations.getOrganisation(1)`.

api/swagger/swagger.yaml
vitorbaptista: It felt redundant to add a description `URL to the Terms and Conditions` or similar...

nightsh: I know, but I added it to most of the fields we had. It seems all the good docs out there have redundant descriptions 😕 
Plus, they look nicer in the docs (depending on the template used). And as a user I guess one would feel safer to get to a docs site where even everything is filled in (we're not there yet, but on our way).

api/swagger/swagger.yaml
nightsh: In the same time, swagger disagrees with reused `operationId`s and outputs red errors in the editor if it finds more of the same.
It also has a point in being able to identify the operations by ID if needed, hence the name. The main idea is, actually, that we should not reuse `operationId`s, as we're not reusing IDs in any of the other specifications we're using (databases, markup etc.)

nightsh: I also like the idea of being semantically nice to write / read as in your example. I just couldn't figure out a way to tackle both issues, and adhering to the Swagger specifications was just higher priority when I operated those.

vitorbaptista:
@nightsh I think it'll be easier to fix the Swagger issues on a separate PR, so merging this now.

nightsh:
👍

PR #82, opened 2016-09-08
[#387] Add entity File that belongs to many Documents

vitorbaptista:
This extracts the attributes related to the actual file location and its text
from the `documents` table to a new `files` table. The reasoning behind it is
that multiple documents can (and do) reference the same file, so this
extraction will allow us to keep a single entry in our database in those cases.

I haven't removed the columns from the `documents` table yet because I want to
populate the `files` table with our current contents before that. After we run
this migration and populate the `files` table, we can add another migration to
remove the then redudant columns.

I also changed to use the "documentcloud_id" instead of its URL so we can
delete files from DocumentCloud as needed (e.g. when they're updated). This
isn't ideal, but we can generate a document's URL based on its ID.

The API changes in this pull request are needed just while we don't finish the
data migration. They will be reverted afterwards.

opentrials/opentrials#387

### DIFF
--- a/api/models/document.js
+++ b/api/models/document.js
@@ -1,5 +1,7 @@
 'use strict';
 
+require('./file');
+
 const bookshelf = require('../../config').bookshelf;
 const BaseModel = require('./base');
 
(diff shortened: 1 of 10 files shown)


api/models/document.js
```diff
@@ -8,10 +10,22 @@ const Document = BaseModel.extend({
...
-    'url',
-    'documentcloud_url',
-    'text',
   ],
+  file: function () {
+    return this.belongsTo('File');
+  },
+  virtuals: {
+    url: function () {
+      console.log('FOOO', this.related('file').toJSON());
```
nightsh: Hey! :-D

nightsh:
Read the whole diff, makes sense, haven't tested it though. If you feel like it needs another run on a different env before merging, do tell and I'll run it locally.

api/models/document.js
vitorbaptista: Ops! Fixed.

vitorbaptista:
@nightsh Thanks, I don't think that's needed in this case :+1:

PR #83, opened 2016-09-08
Fix seeds and make sure Travis catches these erros in the future

vitorbaptista:
The problems happen when we delete/create tables in the wrong order and end up
with a Postgres error. This commit fixes the current seeds and run them twice
on Travis, so those mistakes will end up in a failed build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ before_install:
 before_script:
   - psql -c 'create database opentrials_api_test;' -U postgres
   - npm run migrate
+  # Run seed twice to make sure we're cleaning the DB correctly
+  - npm run seed
   - npm run seed
   - npm run reindex
 script:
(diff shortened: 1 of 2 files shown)


vitorbaptista:
@nightsh Merging this before your review, but please check it whenever you can

nightsh:
Not sure why running the seed twice is needed, but I guess you have seen something I didn't in Travis' logs. I'll keep an eye on it myself 👍

vitorbaptista:
When we run the seeds file, it'll remove all data from the database. However, the order we remove the data is important: we can't remove a `trial` before removing all `trials_interventions` related to it, for example.

When running tests on Travis, the DB is empty initially, so these errors won't appear. However, by running it twice, we make sure that we're deleting the data in the correct order.

PR #84, opened 2016-09-08
[#387] Clean Documents table removing unused fields

vitorbaptista:
I migrated the data manually to the new schema (see
591eba12). Now we're ready to remove the unused fields.

Fixes opentrials/opentrials#387

### DIFF
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -779,6 +779,7 @@ definitions:
     required:
       - name
       - type
+      - url
     properties:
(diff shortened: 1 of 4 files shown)


nightsh:
So instead of having the URLs stored within documents, which might lead to data duplication, this pushes the normalization process further on by adding a relationship with a `files` table containing the link(s) to what used to be considered "documents". If I'm right, then 👍 and I'll have to re-merge the API changes into my PRs.

PR #85, opened 2016-09-14
Update dependencies

vitorbaptista:
I went over all of our dependencies and updated them to their latest version. This isn't what we should usually do, but there're some vulnerability fixes that are good to have (see https://david-dm.org/opentrials/api)

I haven't updated `swagger-hapi` because that breaks our code. We'll need to handle that later.

When merging, please squash the commits.


nightsh:
Working good, tested & ready to meet the `master`'s `HEAD` 👍

PR #86, opened 2016-09-20
Return all trial's sources metadata in trials.sources object

vitorbaptista:
See https://github.com/opentrials/opentrials/issues/327 for the reasoning behind this.

### DIFF
--- a/api/models/publication.js
+++ b/api/models/publication.js
@@ -31,7 +31,7 @@ const Publication = BaseModel.extend({
       id: attributes.id,
       url: attributes.url,
       title: attributes.title,
-      source: this.related('source').toJSON(),
+      source_id: this.attributes.source_id,
     };
 
     return result;
--- a/api/models/record.js
+++ b/api/models/record.js
@@ -40,7 +40,7 @@ const Record = BaseModel.extend({
     const attributes = this.toJSON();
     const result = {
       id: attributes.id,
-      source: this.related('source').toJSON(),
+      source_id: this.attributes.source_id,
     };
 
     if (attributes.url) {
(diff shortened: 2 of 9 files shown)


nightsh:
Some comments:

* Shouldn't we remove the `source_id` from the Trial model?
* There are some schema validation errors (trying now on `/search` endpoint)

api/models/publication.js
```diff
@@ -31,7 +31,7 @@ const Publication = BaseModel.extend({
       id: attributes.id,
       url: attributes.url,
       title: attributes.title,
-      source: this.related('source').toJSON(),
+      source_id: this.attributes.source_id,
```
nightsh: Why not `attributes.source_id`? Same for `record.js`

vitorbaptista: This is a bit messy :/

`attributes` here is `this.toJSON()`. However, it'll only contain the model's visible attributes. `source_id` isn't one of them, so we need to get directly from `this.attributes`

vitorbaptista:
> * Shouldn't we remove the `source_id` from the Trial model?

Not now, as it's important to know which actual source we're displaying info from. We'll end up removing it only when (and if) we start merging attributes from multiple sources.

> * There are some schema validation errors (trying now on /search endpoint)

As we talked, you need to run `npm run reindex`

api/models/publication.js
nightsh: I was expecting something like that, but I thought I'd double check with you before testing.

Why can't we just do `const attributes = this.attributes`? Might be a silly question, but I didn't get to run it yet.

nightsh:
All good, you can merge it as far as I'm concerned 👍

PR #87, opened 2016-09-21
Models for Cochrane reviews integration

georgiana-b:
The schema follows the one discussed in opentrials/opentrials#308.


api/models/risk_of_bias.js
```diff
@@ -0,0 +1,51 @@
...
+    return this.belongsTo('Source');
+  },
+  trial: function() {
+    return this.belongsTo('Trial');
+  },
+  risk_of_bias_criteria: function() {
+    return this.belongsToMany('RiskOfBiasCriteria', 'risk_of_biases_risk_of_bias_criterias',
+      'risk_of_bias_id', 'risk_of_bias_criteria_id').withPivot(['value']);
+  },
+  toJSONSummary: function () {
```
vitorbaptista: Is this used anywhere? If not, please remove it.

api/models/risk_of_bias_criteria.js
```diff
@@ -0,0 +1,34 @@
...
+  hasTimestamps: true,
+  visible: [
+    'id',
+    'name'
+  ],
+  risk_of_bias: function() {
+    return this.belongsToMany('RiskOfBias', 'risk_of_biases_risk_of_bias_criterias',
+      'risk_of_bias_id', 'risk_of_bias_criteria_id').withPivot(['value']);
+  },
+  toJSONSummary: function () {
```
vitorbaptista: Is this used anywhere? If not, please remove it.

migrations/20160912132241_create_risks_of_bias.js
```diff
@@ -0,0 +1,53 @@
...
+    table.uuid('risk_of_bias_id')
+    .references('risk_of_biases.id');
+    table.uuid('risk_of_bias_criteria_id')
+    .references('risk_of_bias_criterias.id');
+
+    table.enu('value', [
+      'yes',
+      'no',
+      'unknown',
+    ]).notNullable();
```
vitorbaptista: I'm not sure where our previous chat about it went. If these are the only 3 valid states, and there's no possibility for a `null`, I'd prefer having it as a nullable boolean instead.

api/models/trial.js
```diff
@@ -106,6 +106,9 @@ const Trial = BaseModel.extend({
   records: function () {
     return this.hasMany('Record');
   },
+  risk_of_biases: function () {
```
vitorbaptista: Rename this to `risks_of_bias`. I know I argued for `risk_of_biases` myself, but that's only for our internal structure. The name of this method will appear to our users, so we should use the correct English.

api/models/risk_of_bias.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const bookshelf = require('../../config').bookshelf;
+const BaseModel = require('./base');
+require('./source');
+require('./trial');
+require('./risk_of_bias_criteria')
+require('../helpers');
```
vitorbaptista: To keep it consistent with the rest of the code, move the `require('./MODEL')` to before the other requires, and remove the unused `require('../helpers')` (see https://github.com/opentrials/api/blob/master/api/models/trial.js for an example of what I mean)

api/models/risk_of_bias.js
```diff
@@ -0,0 +1,51 @@
...
+
+const relatedModels = [
+  'source',
+  'trial',
+  'risk_of_bias_criteria'
+];
```
vitorbaptista: As they're not used, remote the `relatedModels`.

api/models/risk_of_bias.js
```diff
@@ -0,0 +1,51 @@
...
+    'source_url',
+    'created_at',
+    'updated_at'
+  ],
+  source: function () {
+    return this.belongsTo('Source');
+  },
+  trial: function() {
+    return this.belongsTo('Trial');
+  },
```
vitorbaptista: We don't use the `source` or `trial` relationships anywhere yet, so please remove those.

migrations/20160912132241_create_risks_of_bias.js
```diff
@@ -0,0 +1,53 @@
...
+    table.uuid('id').primary();
+
+    table.uuid('trial_id')
+      .notNullable()
+      .references('trials.id');
+    table.text('source_id')
+      .notNullable()
+      .references('sources.id');
+    table.text('source_url')
+      .notNullable();
```
vitorbaptista: Couldn't we receive a risk of bias assessment without a `source_url`? As when we received them as a data donation? If so, this should be `nullable()`

migrations/20160912132241_create_risks_of_bias.js
```diff
@@ -0,0 +1,53 @@
...
+    table.text('study_id')
+      .notNullable();
```
vitorbaptista: What's the `study_id`?

migrations/20160912132241_create_risks_of_bias.js
```diff
@@ -0,0 +1,53 @@
...
+    table.text('name')
+      .notNullable();
+    table.timestamps(true, true);
+  });
+
+  schema.createTable('risk_of_biases_risk_of_bias_criterias', (table) => {
+    table.uuid('risk_of_bias_id')
+    .references('risk_of_biases.id');
+    table.uuid('risk_of_bias_criteria_id')
+    .references('risk_of_bias_criterias.id');
```
vitorbaptista: Please fix the identation here.

api/models/risk_of_bias_criteria.js
```diff
@@ -0,0 +1,34 @@
...
+const relatedModels = [
+  'risk_of_bias',
+];
+
+const RiskOfBiasCriteria = BaseModel.extend({
+  tableName: 'risk_of_bias_criterias',
+  hasTimestamps: true,
+  visible: [
+    'id',
+    'name'
```
vitorbaptista: The `value` from the pivot table should be visible as well

api/models/risk_of_bias_criteria.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const bookshelf = require('../../config').bookshelf;
+const BaseModel = require('./base');
+require('./risk_of_bias');
+
+const relatedModels = [
+  'risk_of_bias',
+];
```
vitorbaptista: We're not using the `relatedModels` here, so please remove it.

api/models/risk_of_bias_criteria.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const bookshelf = require('../../config').bookshelf;
+const BaseModel = require('./base');
+require('./risk_of_bias');
```
vitorbaptista: To keep it consistent with the rest of the code, move the require('./MODEL') to before the other requires (see https://github.com/opentrials/api/blob/master/api/models/trial.js for an example of what I mean)

migrations/20160912132241_create_risks_of_bias.js
georgiana-b: The reasoning behind `source_url` and `study_id` is that we should be able to identify an existing risk of bias uniquely. In order to identify it we have to know which review file it came from and to what study it was attributed. So, `source_url` is a reference to the review file and `study_id` something that identifies the risk of bias attributed to a certain study among other review results present in that review file.

migrations/20160912132241_create_risks_of_bias.js
georgiana-b: I think that might be confusing because `null` means there was no value while `unknown` is a value that reviewers assign when the result is uncertain.
Ex: 
```xml
<QUALITY_ITEM_DATA_ENTRY RESULT="UNKNOWN" STUDY_ID="STD-Baker-2006">
     <DESCRIPTION>
        <P>
            Participants drew a card from an envelope but no details provided regarding the generation of the random sequence or whether cards were shuffled beforehand.
        </P>
    </DESCRIPTION>
</QUALITY_ITEM_DATA_ENTRY>
```

migrations/20160912132241_create_risks_of_bias.js
vitorbaptista: The `study_id` isn't enough to uniquely identify a risk of bias, then? In that case, you should create a unique constraint for `source_url` and `study_id`.

vitorbaptista:
I thought better about this and I agree with you. Leave it as is, with yes,
no and unknown.

seeds/trials.js
```diff
@@ -479,6 +484,55 @@ exports.seed = (knex) => {
...
+  const riskOfBiasCriterias = [
+    {
+      id: '756b06ca-8414-11e6-a6fa-e4b3181a2c8c',
+      name: '3.1',
+      created_at: new Date('2016-01-02'),
+      updated_at: new Date('2016-02-01'),
+    },
+    {
+      id: '7036b016-841c-11e6-a6fa-e4b3181a2c8c',
+      name: '3.2',
```
vitorbaptista: Shouldn't the names be things like `double blinding`?

georgiana-b: I remember we agreed to keep this because the natural language formulation might change more often and we don't want to run migrations every time that happens. We can keep a mapping in the model, but I can also change it to natural language if necessary.

georgiana-b:
@vitorbaptista I noticed the tests check certain use-cases and I realized when I tried to write some tests for these models that it's not clear to me how we plan to use them. Can you give me an example use-case or request-result?

georgiana-b:
Both `npm run test` and `npm run seed` work ok on my local repo and in Victor's clone  so it's unclear why Travis fails. I will investigate this today but please comment if you have any suggestions.

vitorbaptista:
@georgiana-b To fix the Travis build you need to change the https://github.com/opentrials/api/blob/master/test/common.js#L10-L29 file, adding the new tables you created to be cleaned. Ordering is important, as you can't clean a table if there're any other tables referencing it.

migrations/20160912132241_create_risks_of_bias.js
```diff
@@ -0,0 +1,55 @@
...
+    table.primary(['risk_of_bias_id', 'risk_of_bias_criteria_id']);
+  });
+
+  return schema;
+};
+
+exports.down = (knex) => (
+  knex.schema
+    .dropTableIfExists('risk_of_biases_risk_of_bias_criterias')
+    .dropTableIfExists('risk_of_bias_criterias')
```
vitorbaptista: Not sure why dropping the column here, as the `risk_of_biases` table will be removed anyway.

nightsh: Well, I've been able to reproduce the error locally. Did this to dodge the constraint violation. Worked.
(then I looked around a bit, saw the same approach [on `files` table migration](https://github.com/opentrials/api/blob/master/migrations/20160907170328_create_files_belonging_to_many_documents.js#L30))

vitorbaptista: In the `files` table migration, we're removing the foreign key in the `documents` table, not in the `files` table. The error in Travis was caused by https://github.com/opentrials/api/pull/87#issuecomment-249912071, already fixed by @georgiana-b :+1:

api/models/trial.js
```diff
@@ -106,6 +107,9 @@ const Trial = BaseModel.extend({
   records: function () {
     return this.hasMany('Record');
   },
+  risks_of_bias: function () {
+    return this.hasMany('RiskOfBias');
+  },
```
vitorbaptista: Need to add the `risk_of_bias` and `risk_of_bias.risk_of_bias_criteria` to the `relatedModels` here. Also, please write a few tests validating that it's working as expected. They can be simple.

vitorbaptista: Pending...

seeds/trials.js
vitorbaptista: Change the `name` from numbers to actual human-readable names (check `processors/cochrane/extractors.py#extract_review_results` on https://github.com/opentrials/processors/pull/59/files for example of names)

seeds/trials.js
```diff
@@ -117,6 +117,11 @@ exports.seed = (knex) => {
       terms_and_conditions_url: 'http://www.isrctn.com/page/terms',
       type: 'register',
     },
+    cochrane: {
+      id: 'cochrane',
+      name: 'Cochrane Library',
+      type: 'other',
+    },
```
vitorbaptista: Missing `url` and it should be `Cochrane Schizophrenia`  AFAIK

PR #89, opened 2016-09-29
[#418] Limit memory usage when reindexing

vitorbaptista:
NodeJS uses up to 1.5GB by default, but we only have 512MB available in Heroku.

Fixes opentrials/opentrials#418


vitorbaptista:
@nightsh Merging this myself because I have tested it locally and we're already doing the same for the web process, so it works.

nightsh:
😅

PR #91, opened 2016-10-01
[#429] A document must have a "url" or a "file_id", but not both

vitorbaptista:
A document doesn't need to be a file. It can be a webpage, for example.

Related to https://github.com/opentrials/processors/pull/66

Fixes opentrials/opentrials#429


nightsh:
It's good to go, even if my review is post-mortem 😀

PR #98, opened 2016-10-06
[#287] More details on the search query string

nightsh:
* Mention that `q` is an Elasticsearch query string

Fixes opentrials/opentrials#287


nightsh:
@vitorbaptista kudos for the completions, let's merge this, shall we?

PR #99, opened 2016-10-06
[#474] Improve /search endpoint performance

vitorbaptista:
Using `res.json()` calls to `JSON.stringify()` on the response. As we're
getting the response from ElasticSearch, all fields are already converted to
JSON (there're no Date classes, for example), so there's no need to call
`JSON.stringify()`.

That reduces the response time in about 50%. You can check by using:
  ab -n 1000 -c 50 http://localhost:10010/v1/search

With and without this patch.

opentrials/opentrials#474


nightsh:
👍

PR #100, opened 2016-10-13
[#486] Ignore ICTRP when calculating discrepancies

nightsh:
Fixes opentrials/opentrials#486


nightsh:
Testing this is a bit silly now, suggestions are welcome. I tried to chain a `forEach` on a list of ignored sources, but didn't get the chaining right. It works as it is, but it could be improved.

PR #102, opened 2016-10-20
[#499] Expose FDA applications, approvals and documents

georgiana-b:
opentrials/opentrials#499


api/controllers/fda_applications.js
```diff
@@ -0,0 +1,51 @@
...
+
+function getFdaApplications(req, res) {
+  const params = req.swagger.params;
+  const page = params.page.value;
+  const perPage = params.per_page.value;
+
+  return FdaApplication.fetchPage({
+      page: page,
+      pageSize: perPage,
+      withRelated: ['fda_approvals', 'organisation'],
```
vitorbaptista: Use the `FdaApplication` defined in the model instead

api/controllers/fda_applications.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const FdaApplication = require('../models/fda_application');
+
+function getFdaApplication(req, res) {
+  const id = req.swagger.params.id.value;
+
+  return new FdaApplication({ id: id }).fetch({
+      withRelated: ['fda_approvals', 'organisation'],
```
vitorbaptista: Use the `FdaApplication` defined in the model instead

api/controllers/fda_applications.js
```diff
@@ -0,0 +1,51 @@
...
+      if (fda_application) {
+        res.json(fda_application);
+      } else {
+        res.status(404);
+        res.finish();
+      }
+    });
+}
+
+function getFdaApplications(req, res) {
```
vitorbaptista: To keep naming consistenty with https://github.com/opentrials/api/blob/master/api/controllers/sources.js, name it `listFdaApplications`

api/swagger/swagger.yaml
```diff
@@ -974,6 +1030,59 @@ definitions:
       name:
         type: string
 
+  FdaApplications:
```
vitorbaptista: Rename to `FdaApplicationList` (to keep consistency with `SourceList`)

test/api/controllers/fda_applications.js
```diff
@@ -0,0 +1,66 @@
...
+    it('returns 404 if there\'s no FDA application with the received ID', () => (
+      server.inject('/v1/fda_applications/BADFDAID')
+        .then((response) => {
+          response.statusCode.should.equal(404);
+        })
+    ));
+
+    it('returns the FDA application', () => (
+      factory.create('fda_application').then((model) => {
+        return server.inject('/v1/fda_applications/' + model.attributes.id)
```
vitorbaptista: Prefer ES6 style: \`/v1/fda_applications/${model.attributes.id}\`

test/api/controllers/fda_applications.js
```diff
@@ -0,0 +1,66 @@
...
+          .then((response) => {
+            response.statusCode.should.equal(200);
+
+            const expectedResult = JSON.parse(JSON.stringify(model.toJSON()));
+            const result = JSON.parse(response.result);
```
vitorbaptista: There's a global method (ugly, I know :sad:) named `toJSON()` that you can use here, as `const expectedResult = toJSON(model)` (I think you don't need to do `toJSON(model.toJSON())`, but I haven't tested)

Also, I don't think you need to parse the `response.result`, as it should already be handled by Hapi.

test/api/controllers/fda_applications.js
```diff
@@ -0,0 +1,66 @@
...
+  describe('GET /v1/fda_applications', () => {
+    it('returns the FDA applications in pages', () => (
+      factory.create('fda_application').then((model) => {
+        return server.inject('/v1/fda_applications')
+          .then((response) => {
+            response.statusCode.should.equal(200);
+
+            const expectedResult = {
+              total_count: 1,
+              items: [JSON.parse(JSON.stringify(model.toJSON()))],
```
vitorbaptista: Same as above, I think `toJSON(model)` would be enough.

test/api/controllers/fda_applications.js
```diff
@@ -0,0 +1,66 @@
...
+      factory.create('fda_application').then((model) => {
+        return server.inject('/v1/fda_applications?page=2')
+          .then((response) => {
+            response.statusCode.should.equal(200);
+
+            const expectedResult = {
+              total_count: 1,
+              items: [],
+            }
+            const result = JSON.parse(response.result);
```
vitorbaptista: As above, I guess there's no need to parse the result

test/factory.js
```diff
@@ -227,4 +229,28 @@ factory.define('sourceRelatedToSeveralRecords', Source, {
...
+  drug_name: 'Healer',
+  active_ingredients: 'healing',
+  organisation_id: factory.assoc('organisation', 'id'),
+}, {
+  afterCreate: (fdaApplication, attrs, callback) => {
+    factory.create('fda_approval', { fda_application_id: fdaApplication.id })
+    .then(() => new FdaApplication({ id: fdaApplication.id })
+      .fetch({ withRelated: ['fda_approvals', 'organisation'] }))
+    .then((instance) => callback(null, instance))
+    .catch((err) => callback(err));
```
vitorbaptista: Fix indentation and use the model's `FdaApplication.relatedModels`, so we avoid repetitions

test/factory.js
```diff
@@ -227,4 +229,28 @@ factory.define('sourceRelatedToSeveralRecords', Source, {
...
+  },
+});
+
+factory.define('fda_approval', FdaApproval, {
+  id: () => uuid.v1(),
+  supplement_number: 0,
```
vitorbaptista: If I remember correctly, there's a uniqueness constraint on `supplement_number` and `fda_application_id`, so you should use `factory.sequence()` here.

api/controllers/fda_applications.js
```diff
@@ -0,0 +1,51 @@
...
+function getFdaApplication(req, res) {
+  const id = req.swagger.params.id.value;
+
+  return new FdaApplication({ id: id }).fetch({
+      withRelated: ['fda_approvals', 'organisation'],
+    })
+    .catch((err) => {
+      res.finish();
+      throw err;
+    })
```
vitorbaptista: The `catch` should be the last thing we do in the promise, as a fallback for any error that happened (e.g. if `res.json()` on line 17 throws an error, this code will hang).

api/models/fda_application.js
```diff
@@ -0,0 +1,32 @@
...
+    'fda_approvals',
+    'organisation',
+  ],
+  fda_approvals: function () {
+    return this.hasMany('FdaApproval');
+  },
+  organisation: function () {
+    return this.belongsTo('Organisation');
+  }
+},{
```
vitorbaptista: Small styling: `}, {` instead of `},{` (notice the extra whitespace)

api/controllers/fda_applications.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const FdaApplication = require('../models/fda_application');
```
vitorbaptista: Use `FDAApplication` instead, as `FDA` is an acronym (not only here, but throughout the PR)

api/models/document.js
```diff
@@ -2,40 +2,79 @@
...
+    return attributes
+  },
+  file: function () {
+    return this.belongsTo('File');
+  },
+  trials: function () {
+    return this.belongsToMany('Trial', 'trials_documents');
+  },
+  toJSONSummary: function () {
+    const attributes = this.attributes;
```
georgiana-b: I used `this.attributes` here because I needed access to `source_url` which is not visible and thus not available through `toJSON()`.

test/api/models/document.js
```diff
@@ -43,8 +55,30 @@ describe('Document', () => {
...
+    it('returns summarized version of file', () => {
+      return factory.create('file')
+        .then((file) => factory.create('documentWithFile', { file_id: file.attributes.id }))
+        .then((doc) => new Document({ id: doc.attributes.id }).fetch({ withRelated: ['file'] }))
+        .then((doc) => doc.toJSON().file.should.deepEqual(doc.related('file').toJSONSummary()));
+    });
+    it('returns empty object if it lacks file', () => {
+      return factory.create('document', {file_id: undefined})
+        .then((doc) => new Document({ id: doc.attributes.id }).fetch({ withRelated: ['file'] }))
+        .then((doc) => doc.toJSON().file.should.deepEqual({}));
```
georgiana-b: I should've used `equal()` instead of `deepEqual()` but for some reason that raised: `AssertionError: expected Object {} to be Object {}` :smile:.

vitorbaptista: Not related to your comment, but it's usually better to use `should(foo).equal(bar)` instead of `foo.should.equal(bar)`, because if `foo` is `undefined`, this will throw a kind of cryptic error message (`undefined has no method 'should'` or similar.)

Also, if a `doc` has no `file`, it would be better if `doc.file == undefined`, so the API users won't need to test against `{}` as well.

georgiana-b:
@vitorbaptista Turns out the issue we had on `/documents` with `file` relation being an empty object comes from `fetchPage()` with uses `fetchAll()` which has [this known issue](https://github.com/tgriesser/bookshelf/issues/753).

So, the patch you made on `Document` would solve it but we're going to have the same problem anywhere we use `fetchAll` (or other collection methods) and there is a `1-to-1` relation.
Ex: In [`listFDAApplications`](https://github.com/georgiana-b/api/blob/1958e263954afb7d1ff6ffad4536978134b3fc26/api/controllers/fda_applications.js#L30). 

Bookshelf already [has an issue open particularly for this](https://github.com/tgriesser/bookshelf/issues/1299) which has been labeled as `More information required`. Maybe we should share our case with them? 

Anyway, while we wait for someone to solve it (or make a contribution), we could simply use [the option `model.toJSON({ omitNew: true})`](https://github.com/tgriesser/bookshelf/pull/1286) which will exclude attributes that are empty objects. 

Unfortunately I couldn't get it to work, hopefully because I'm misusing it. Maybe you'll have more success.
If you won't, we're in for even more patching. :confused:

vitorbaptista:
@georgiana-b I couldn't use `{omitNew: true}` either. However, I refactored the code to move the logic to the `BaseModel`, so we keep the code a bit more dry (see https://github.com/opentrials/api/pull/102/commits/62f04e740586f99aa01aecda56d72b2a1bc9c8fb).

I also renamed a few variables named `document`, as they're reserved words in JS (https://github.com/opentrials/api/pull/102/commits/320f62af1c4370f19ce0c448c022c8b8a2cc8d9a). There might be others I didn't see.

PR #105, opened 2016-10-25
Fix documents uniqueness constraints

vitorbaptista:
We should never alter migrations, but I realized the ones we have now will not
work on production. The problem is that our uniqueness constraints are
different for FDA documents and non-FDA documents. This is a very clear signal
that we should have a separate FDA document entity (see
opentrials/opentrials#521). We don't have it now, so this is the best we can
do.


pwalsh:
"We should never alter migrations". Except, of course, when we do :). Love the pragmatism.

PR #106, opened 2016-10-27
[#425] Return all sources for a trial

nightsh:
Warning: this introduces a breaking change on the returned sources list for a trial, which is now an array on unique elements as opposed to a hash keyed by `source_id`.

The `sources` virtual now also collects sources from `documents` and adds them to the list.

Swagger definitions and tests modified accordingly.

**Note:** I'm quite happy for moving towards lists of things that are actually sets/arrays, which is likely more predictable from an user's PoV. I wouldn't really expect `trial.sources` to give me back an object :)

**Note 2:** Special attention to be paid by reviewer to the other potentially overlooked usages of `sources` ex-object, now array, throughout the code. If something is missing, it won't be observable from the PR diff alone.

### DIFF
--- a/api/controllers/documents.js
+++ b/api/controllers/documents.js
@@ -33,9 +33,9 @@ function listDocuments(req, res) {
       withRelated: Document.relatedModels,
     })
     .then((documents) => {
-      let response = {
+      const response = {
         total_count: documents.pagination.rowCount,
-        items: documents.models,
+        items: documents.models.map((m) => m.toJSONSummary()),
       }
       res.json(response);
     })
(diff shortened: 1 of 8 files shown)


vitorbaptista:
@nightsh Although I agree that `trial.sources` feel like it should be an array, it ends up complicating the client code, as we need to get the source by ID in a few places (not only the `records`, but also the `identifiers`, missed in your https://github.com/opentrials/opentrials/pull/528 PR).

We shouldn't design the API only to make it easier only for the Explorer, as we don't plan on being its sole users. However, I do think that knowing the details of a trial's source is a pattern common enough for us to optimize for it (i.e. keep `trial.sources` as an object keyed by `id`).

PR #107, opened 2016-10-27
[#499] Remove extra fields from DocumentSummary and allow null source_id

vitorbaptista:
We have some documents without `source_id` that came from data contributions.
We'll change them to include the original sources, but while we don't do that,
we need to allow nullable fields.

opentrials/opentrials#499

Related to opentrials/opentrials#530


vitorbaptista:
@georgiana-b I'm merging this because it's broken on production. Please do review it regardless.

PR #111, opened 2016-11-01
[#542] Remove the schemes accepted in the Swagger definition

vitorbaptista:
The current problem is that the swagger clients are trying to load the API
using HTTP, instead of HTTPS. If that's on client-side, as it is on the swagger
docs page, the browser blocks those requests because they're "insecure" (mixed
content).

We could change the scheme to be HTTPS-only, but then it'll break local
development, as we don't use HTTPS while developing it on our local machines.

By removing the schemes, Swagger will use the same scheme as the one used to
load the Swagger definition. In other words, if you load the `swagger.yaml` via
HTTPS (like in production), it'll use HTTPS; if you load it via HTTP (like in
development), it'll use HTTP.

Fixes opentrials/opentrials#542

### DIFF
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -3,9 +3,6 @@ info:
   version: "0.0.1"
   title: OpenTrials API
 basePath: /v1
-schemes:
-  - http
-  - https
 consumes:
   - application/json
 produces:
--- a/app.json
+++ b/app.json
@@ -16,7 +16,7 @@
       "options": {
         "version": "9.4"
       }
-    }
+    },
     {
       "plan": "bonsai:sandbox",
       "options": {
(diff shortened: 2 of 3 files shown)


pwalsh:
can't we set scheme to be an env var, and also enforce https in production? For similar reasons I did something like that [here](https://github.com/okfn/opendatacensus/blob/master/census/config/index.js#L37)

vitorbaptista:
@pwalsh We could, but for that we'd need to create a build process for the `swagger.yaml` file, as it's currently written manually and exposed as is.

The worst-case scenario with this PR is if a user loads the swagger definition via HTTP. CloudFlare will automatically redirect that to HTTPS, but I'm not sure if the swagger client is smart enough to get the scheme from the redirected URL, and not the one it received. I'll test that when this is merged (i.e. soon).

PR #112, opened 2016-11-01
[#543] Fix bug with `trial.sources` when its documents have no source

vitorbaptista:
`document.source_id` is nullable for now. It'll change in the future (see
opentrials/opentrials#530), but for now we need to solve this bug.

Fixes opentrials/opentrials#543


vitorbaptista:
@nightsh Merging this because it's a quite bad issue in production, but please do review it.

nightsh:
Oh, good catch! Kudos 👍

PR #114, opened 2016-11-02
[#544] Add Risk of Bias sources to trial

nightsh:
Fixes opentrials/opentrials#544

Fix for getting the Risk of Bias sources into the `trial.sources` virtual.

### DIFF
--- a/api/models/trial.js
+++ b/api/models/trial.js
@@ -31,6 +31,7 @@ const relatedModels = [
   'documents.fda_approval',
   'documents.fda_approval.fda_application',
   'risks_of_bias',
+  'risks_of_bias.source',
   'risks_of_bias.risk_of_bias_criteria',
 ];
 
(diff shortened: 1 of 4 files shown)


api/models/trial.js
```diff
@@ -133,14 +133,25 @@ const Trial = BaseModel.extend({
...
                                  .map((record) => record.source);
+      const robSources = this.related('risks_of_bias')
+                             .toJSON()
+                             .map((rob) => {
+                               return {
+                                 id: rob.source_id,
+                                 name: "Cochrane Schizophrenia group",
+                                 source_url: rob.source_url,
+                               };
+                             });
```
vitorbaptista: We do have the source in our database, see https://github.com/opentrials/processors/blob/master/processors/cochrane_reviews/extractors.py#L45-L51. So this is the same as the other entities (`record`, `documents`, etc.).

While you're there, as you changed the `documentsSources` anyway, please change to use `doc` instead of `document`, as `document` is a reserved keyword in JavaScript.

nightsh: Ah, you're right! I completely overlooked that. Need more ☕️

nightsh: @vitorbaptista check it out now

vitorbaptista:
@nightsh Could you rebase with master and fix the merge conflicts? Other than that, this is good to go, right?

nightsh:
@vitorbaptista I'm afraid you have to press some buttons on your side, since you requested changes :)

vitorbaptista:
@nightsh Just missing some tests and good to go. There're tests for the other entities, so this should be simply a matter of copy/pasting.

PR #115, opened 2016-11-03
Add /v1/search/fda_documents endpoint

vitorbaptista:
See the commit messages for more background on what was done and why. This fixes opentrials/opentrials#500.

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -40,6 +40,102 @@ function searchTrials(req, res) {
     });
 }
 
+function _convertFDADocumentsElasticSearchResult(esResult) {
+  return {
+    total_count: esResult.hits.total,
+    items: esResult.hits.hits.map((hit) => {
+      const doc = hit._source;
+
+      if (hit.inner_hits.page !== undefined) {
+        const pages = hit.inner_hits.page.hits.hits;
+        doc.file.pages = pages.map((page) => ({
+          text: page._source.text,
+          num: page._source.num,
+        }));
+      }
+
+      return doc;
+    }),
+  };
+}
+
+function searchFDADocuments(req, res) {
+  const params = req.swagger.params;
+  const page = params.page.value;
+  const perPage = params.per_page.value;
+  const queryString = params.q.value;
+
+  let documentQuery;
+  let pageQuery;
+  if (queryString === undefined) {
+    documentQuery = {
+      match_all: {},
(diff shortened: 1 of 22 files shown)


api/controllers/search.js
```diff
@@ -40,6 +40,102 @@ function searchTrials(req, res) {
...
+      match_all: {},
+    };
+    pageQuery = {
+      match_all: {},
+    };
+  } else {
+    const queryStringQuery = {
+      query: queryString,
+      default_operator: 'AND',
+    }
```
nightsh: Missing semicolon here, we need to fix that linter :)

api/models/fda_application.js
```diff
@@ -30,6 +30,9 @@ const FDAApplication = BaseModel.extend({
     url: function () {
       return helpers.urlFor(this);
     },
+    type: function () {
+      return this.id.match(/[A-Z]+/gi)[0];
```
nightsh: I'm not sure I understand this. Is it safe?

migrations/20161026164912_alter_files_text_to_array_and_rename_to_pages.js
```diff
@@ -0,0 +1,27 @@
...
+exports.up = (knex) => (
+  knex.schema
+    .raw(`
+      ALTER TABLE files
+        RENAME COLUMN text TO pages
+    `)
+    .raw(`
+      ALTER TABLE files
+        ALTER COLUMN pages TYPE text[]
+        USING array[pages]::text[]
```
nightsh: Interesting, I didn't know we could do this with postgres!

api/controllers/search.js
vitorbaptista: Definitely!

api/models/fda_application.js
vitorbaptista: The IDs use the format `XXX000000` (e.g. `NDA00000`, `ANDA00000`), so this will return the first group of non-numeric values there.

You're right to be concerned, though. I added a new check in case the id doesn't start with letters, just to be safer :+1:

migrations/20161026164912_alter_files_text_to_array_and_rename_to_pages.js
vitorbaptista: Me neither :smile:

PR #116, opened 2016-11-08
Allow running individual indexers

vitorbaptista:
You can run, for example, `npm run reindex trials fdaDocuments` to reindex only
these two indexes. Now that we're indexing FDA Documents as well, the indexing
process can take a long time, so it's useful to be able to only index the few
that changed.


georgiana-b:
I reviewed the code and tried it locally. All good :+1:

PR #117, opened 2016-11-08
[#555] Fix FDA Docs indexer when working on files without pages

vitorbaptista:
Fixes opentrials/opentrials#555


vitorbaptista:
@georgiana-b had to merge because it's needed in production. Please still review when you can.

georgiana-b:
@vitorbaptista  I checked the code and run the command again. It's fine now.

PR #119, opened 2016-11-10
FDA Document search API tweaks for the search UI

vitorbaptista:
See the commit messages.

opentrials/opentrials#501


georgiana-b:
Nice touch with the highlights! I reviewed the code and tested most important functionality but only on seed data. It looks good :+1: .
I will test on production data too a bit later and open new issues if I find anything.

PR #121, opened 2016-11-14
Add FDA Application and approval to trial.documents and add public_title to Trial JSON Summary

vitorbaptista:


nightsh:
Awesomeness, bring it on! ;)
(looks good, but I haven't tested it)

PR #122, opened 2016-11-15
Fix FDA search bug on documents with empty pages

vitorbaptista:
When a document has an empty page (i.e. `""`), ElasticSearch doesn't return the `highlights` attribute. In that case, we need to use the text from the page's `_source` instead.


georgiana-b:
Seems fine.

PR #123, opened 2016-11-17
Add correct trial.completion_date mapping

vitorbaptista:
Without this, ElasticSearch uses a different mapping that requires dates on the
format 2016-01-01 (don't accept dates like 2016-1-1). This is different from
what we have on `trials.registration_date`, so is surprising to our users.


nightsh:
On it :)

nightsh:
All good!

PR #124, opened 2016-11-25
Sort FDA documents by application and date

vitorbaptista:
The results from the same application should be together, ordered by date. As
we're sorting descending by application number, the newer applications (with
higher numbers) will appear first.


vitorbaptista:
This is ready for review. You can compare the sorting from https://ot-api-next-pr-124.herokuapp.com/ with production

georgiana-b:
Verified and reviewed. Good to go!

PR #125, opened 2016-11-29
Improve documentation to support local deploy

felipevieira:
As a beginner contributor I've faced some minor issues when setting up the project for the first time in my local machine. These were the adjustments I had to do in order to be able to run the project properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ The OpenTrials API service.
 2. Run `npm install`;
 3. Run `npm run migrate`;
 4. (Optional) If you want, you can add some seed data using `npm run seed`;
-5. Run `npm run reindex`;
+5. (If you ran 4) Run `npm run reindex`;
 
 After the install and migrations ran successfully, you can run `npm run dev` to
 run the project. If you haven't changed the default `PORT`, it should be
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ The OpenTrials API service.
 2. Run `npm install`;
 3. Run `npm run migrate`;
 4. (Optional) If you want, you can add some seed data using `npm run seed`;
-5. Run `npm run reindex`;
+5. (If you ran 4) Run `npm run reindex`;
```
vitorbaptista: I suspect that, if the user haven't added seed data, she'll have data in her local DB (e.g. loaded from a dump), so would need to run reindex anyway. Considering this, I prefer to keep step `5` as is.

felipevieira: yeah I was just following the exact to-do described on the readme, so in this process the database was completely clear, causing an error when runnning the reindex, but if you consider that users can do any dumping processes between on any point of the setup, I agree with you and would mantain as it is

felipevieira: I'm not completely sure if our endpoints would fail if there's no indexes on ES, but +1 for avoiding exception from reindex if there's no data on database

felipevieira: anyway, I've just removed the constraint I've added since it's not a documentation issue

vitorbaptista: You're right: it won't. Created an issue for the reason our `npm run reindex` fails with an empty DB. It's a bug on BookshelfJS: https://github.com/tgriesser/bookshelf/issues/1461

PR #126, opened 2016-11-30
[#575] Remove stats endpoint

georgiana-b:
opentrials/opentrials#575


nightsh:
On it together with opentrials/opentrials#579

PR #127, opened 2016-12-04
[#586] Fix wrong dates due to offset of time

georgiana-b:
opentrials/opentrials#686

After a charming journey among the horrors of JS date handling, this is the best thing I could come up with.
Imho expressed in [this node-pg-types issue](https://github.com/brianc/node-pg-types/issues/50), the `Date`-only fields should **not** have any time information attached to them. However, seems that most JS developers (including the ones at `node-elasticsearch`) don't share my opinion so I had to settle with this for consistency.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -63,6 +63,28 @@ bookshelf.plugin('virtuals');
 bookshelf.plugin('pagination');
 config.bookshelf = bookshelf;
 
+// Overwrite node-pg-types date parser. See https://github.com/tgriesser/knex/issues/1750
+const pgTypes = require('pg').types;
+
+function parseDate(val) {
+  return new Date(Date.parse(val));
+}
+function parseDateArray(value) {
+  if (!value) { return null; }
+
+  const p = pgTypes.arrayParser.create(value, (entry) => {
+    let parsedEntry = entry;
+    if (entry !== null) {
+      parsedEntry = parseDate(entry);
+    }
+    return parsedEntry;
+  });
+
+  return p.parse();
+}
+pgTypes.setTypeParser(1082, parseDate);
+pgTypes.setTypeParser(1182, parseDateArray);
+
 // ElasticSearch
 const elasticsearchConfig = {
   host: process.env.ELASTICSEARCH_URL,
--- a/config/index.js
+++ b/config/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 require('dotenv').config();
+require('./pg_types');
 
 if (!process.env.ELASTICSEARCH_URL) {
   // Fallback to BONSAI_URL if ELASTICSEARCH_URL isn't set.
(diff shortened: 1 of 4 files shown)


config/index.js
```diff
@@ -63,6 +63,28 @@ bookshelf.plugin('virtuals');
 bookshelf.plugin('pagination');
 config.bookshelf = bookshelf;
 
+// Overwrite node-pg-types date parser. See https://github.com/tgriesser/knex/issues/1750
+const pgTypes = require('pg').types;
+
+function parseDate(val) {
+  return new Date(Date.parse(val));
```
vitorbaptista: Is it possible for `parseDate()` to be called with null values as well? If so, it needs to handle this case, similarly to `parseDateArray`

config/index.js
```diff
@@ -63,6 +63,28 @@ bookshelf.plugin('virtuals');
...
+    if (entry !== null) {
+      parsedEntry = parseDate(entry);
+    }
+    return parsedEntry;
+  });
+
+  return p.parse();
+}
+pgTypes.setTypeParser(1082, parseDate);
+pgTypes.setTypeParser(1182, parseDateArray);
```
vitorbaptista: When you move this to a separate file, please remove these magic numbers `1082` and `1182`, using a constant like `DATE_OID`, so it gets more readable.

test/config/pg_types.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('returns null when value is null', () => {
+    let testDate = null;
+    should.equal(pgTypes.getTypeParser(DATE_OID)(testDate), null);
+  });
+
+  it('raises when value is invalid', () => {
+    let testDate = 'not a date';
+    pgTypes.getTypeParser(DATE_OID).bind(null, testDate).should.throw(/^Invalid Date/);
```
georgiana-b: I used `.bind()` here because if I just do `pgTypes.getTypeParser(DATE_OID)(testDate).should.throw(/^Invalid Date/)` the tests fail with this error:

```
  1) parseDate raises when value is invalid:
     Error: Invalid Date
      at parseDate (config/pg_types.js:9:650)
      at Context.it (test/config/pg_types.js:25:36)
```

georgiana-b:
@nightsh Please take a look at this and approve if you think it's ok.

config/pg_types.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function parseDateArray(value) {
+  if (!value) {
+    return null;
+  }
+
+  const p = pgTypes.arrayParser.create(value, (entry) => {
+    let parsedEntry = entry;
+    if (String(entry).toLowerCase() !== String(null)) {
```
nightsh: To clarify why is this here: `null` value only enters the function as the string `'null'`. This condition works with either version of it.

test/config/pg_types.js
```diff
@@ -0,0 +1,55 @@
...
+    ]);
+  });
+
+  it('returns null when value is null', () => {
+    let testArray = null;
+    should(parseDateArray(testArray)).equal(null);
+  });
+
+  it('returns empty array when there are no values', () => {
+    let testArray = `{}`;
```
nightsh: I would've named this `testPgArray`, just to avoid any confusions with "regular" arrays delimited by square brackets.

test/config/pg_types.js
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('returns null when value is null', () => {
+    let testDate = null;
+    should(parseDate(testDate)).equal(null);
+  });
+
+});
+
+describe('parseDateArray', () => {
```
nightsh: `it('calls parseDate() on single array items')`

For when you pass it an array of a single item. So we'd avoid creating an interval (unless we really want to).

config/pg_types.js
```diff
@@ -0,0 +1,36 @@
...
+      parsedEntry = parseDate(entry);
+    } else {
+      parsedEntry = null;
+    }
+    return parsedEntry;
+  });
+
+  return p.parse();
```
nightsh: A bit unclear what `p` means, unless you go back up to see it's a parser. Might be worth renaming it to `parser`.

test/config/pg_types.js
georgiana-b: An array of dates does not represent an interval in this case. It's just a normal array with date values.

PR #131, opened 2016-12-14
[#610] Add is_primary to record

georgiana-b:
opentrials/opentrials#610

### DIFF
--- a/migrations/20161213144433_add_is_primary_to_records.js
+++ b/migrations/20161213144433_add_is_primary_to_records.js
@@ -4,7 +4,7 @@ exports.up = (knex) => (
   knex.schema
     .table('records', (table) => {
       table.boolean('is_primary')
-        .nullable();
+        .defaultTo(false);
     })
 );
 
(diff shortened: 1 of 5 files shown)


migrations/20161213144433_add_is_primary_to_records.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema
+    .table('records', (table) => {
+      table.boolean('is_primary')
+        .nullable();
```
nightsh: Shouldn't this default to `false`, instead of being nullable?

seeds/trials.js
```diff
@@ -468,6 +469,7 @@ exports.seed = (knex) => {
       trial_id: trials[1].id,
       source_id: sources.isrctn.id,
       source_url: 'http://www.isrctn.com/ISRCTN11631712',
+      is_primary: null,
```
nightsh: Why not use `true`/`false` instead of `true`/`null`?

georgiana-b:
@nightsh  I added a default to `false` for `is_primary` as you mentioned and removed the logic to only expose the field when its value is `true`, so now it will show in the payload for all `records` whether it's `true` or `false`.

> Another question: do we need the is_primary field in the ES indexer?

As far as I've seen we don't have an index dedicated to `records`. We only index `trials` and `discrepancies` so there's no room for `is_primary`.

PR #134, opened 2016-12-20
Adjust the garbage collector to reduce memory footprint

nightsh:
Attempting to reduce the issues we have with memory load on the API. Step one is to slightly adjust the way Node's GC works and observe if there is any impact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "migrate": "knex migrate:latest",
     "seed": "knex seed:run",
-    "start": "node --optimize_for_size --max_old_space_size=460 --gc_interval=100 server.js",
+    "start": "node --optimize_for_size --max_old_space_size=400 --gc_interval=100 server.js",
     "dev": "nodemon server.js",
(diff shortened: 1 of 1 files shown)


nightsh:
Oh, can I do that? I didn't know, honestly.

Do you know whether the main idea is to leave it on heroku, or switch to our own infra hosting instead? Not sure what are the next steps regarding this, therefore I don't know how much should I research heroku-based optimizations vs more subtle ways of improving performance.

pwalsh:
it will move to our own hosting as soon as @vitorbaptista is back

pwalsh:
and I don't think you should investigate performance optimisations at all. if you have a memory problem, tell me and we'll increase the resources on heroku in the short term

nightsh:
Oh, okay then, feel free to increase the memory then, it's hitting memory quota multiple times a day, e.g.

> Dec 19 23:34:51 opentrials-api heroku/web.1: Error R14 (Memory quota exceeded) 

And we can probably kill this PR afterwards.

pwalsh:
@nightsh you can kill the PR.

PR #136, opened 2017-01-17
[WIP] [#353] Alters the schema of the database and updates the seed

arthurSena:
We modified the migrations and seeds folders in order to alter the 'study_phase' column's type. 
This was done due the solution of  [issue#353](https://github.com/opentrials/opentrials/issues/353) from Opentrials repository.

### DIFF
--- a/migrations/20160216120159_create_initial_schema.js
+++ b/migrations/20160216120159_create_initial_schema.js
@@ -51,7 +51,7 @@ exports.up = (knex) => {
       .notNullable();
     table.text('study_design')
       .notNullable();
-    table.text('study_phase')
+    table.jsonb('study_phase')
       .notNullable();
 
     table.jsonb('primary_outcomes')
--- a/migrations/20160216120159_create_initial_schema.js
+++ b/migrations/20160216120159_create_initial_schema.js
@@ -51,7 +51,7 @@ exports.up = (knex) => {
       .notNullable();
     table.text('study_design')
       .notNullable();
-    table.jsonb('study_phase')
+    table.specificType('study_phase', 'text[]')
       .notNullable();
 
     table.jsonb('primary_outcomes')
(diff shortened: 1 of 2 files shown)


migrations/20160216120159_create_initial_schema.js
```diff
@@ -51,7 +51,7 @@ exports.up = (knex) => {
       .notNullable();
     table.text('study_design')
       .notNullable();
-    table.text('study_phase')
+    table.jsonb('study_phase')
```
vitorbaptista: Why you're using `jsonb` instead of an actual array type (maybe even an array of enum)?

arthurSena: At a first glance at Knex documentation, we didn't see any mention about array types and observing
other tables in the database we opted to go with the jsonb route.
With your suggestion, we went back to the documentation and found the 'specificType' function that allows us to force the array type on
a column.
However, for an array of enum we would have to dive into using Raw SQL object as the enum function 
from Knex creates a constraint instead of an actual enum.
With this in mind, we think that the array of strings approach is better. We would like to know your thoughts on it.

vitorbaptista: Just saw your response. I agree, let's keep it as `text[]`.

vitorbaptista:
@arthurSena Sorry, I was merging this after merging https://github.com/opentrials/processors/pull/103 but just noticed that you altered the original migration.

Migrations should never be altered, as those changes would only propagate if you created the DB again from scratch and ran all migrations. You always have to create new migrations to change the DB schema from whatever it is at the time to what you want. With knex, you create new migrations using `knex migrate:make MIGRATION_NAME`.

As I already merged the processor which requires this DB change (my mistake), this change needs to go in relatively fast, otherwise any processor that runs will break. I went ahead and wrote a PR with the migration at https://github.com/opentrials/api/pull/139. Please take a look for reference.

PR #137, opened 2017-01-23
[#467] Manage created_at and updated_at at db level

georgiana-b:
opentrials/opentrials#467

Currently `updated_at` and `created_at` fields are of type `TIMESTAMP WITH TIME ZONE`. Until now, we [made sure in the code that the timezone will be UTC](https://github.com/opentrials/processors/blob/master/processors/base/writers/condition.py#L33). However, after this change, if we keep the timezone, we have to [tell PostgreSQL to use UTC by editing `postgresql.conf`](https://www.postgresql.org/docs/9.1/static/runtime-config-client.html#GUC-TIMEZONE).

### DIFF
--- a/migrations/20170122130426_set_default_for_created_at_and_updated_at.js
+++ b/migrations/20170122130426_set_default_for_created_at_and_updated_at.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const Promise = require('bluebird');
+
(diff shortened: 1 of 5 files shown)


georgiana-b:
@vitorbaptista I noticed our `updated_at` and `created_at` have type `TIMESTAMP WITH TIME ZONE`. Is that intentional? If so, why do we want them with timezone?

PR #139, opened 2017-01-25
[#353] Alter trials and records' study_phase to text[]

vitorbaptista:
Fixes opentrials/opentrials#353


vitorbaptista:
Merging this in because it's required for running any processor that uses `study_phase`.

PR #140, opened 2017-01-27
Add `is_registered` boolean to identify unregistered trials

nightsh:
opentrials/opentrials#655

### DIFF
--- a/api/models/trial.js
+++ b/api/models/trial.js
@@ -161,6 +161,7 @@ const Trial = BaseModel.extend({
           data[source.id] = {
             id: source.id,
             name: source.name,
+            type: source.type,
             source_url: source.source_url,
           };
         }
@@ -209,6 +210,12 @@ const Trial = BaseModel.extend({
       }
 
       return discrepancies;
+    },
+    is_registered() {
+      // Be aware that this depends on `source_id` to be accurate
+      // So if a trial has a `nct` source but a `pubmed` source_id, the bug is elsewhere
+      const source = this.sources[this.attributes.source_id];
+      return source.type === 'register';
     },
   },
 }, {
--- a/api/models/trial.js
+++ b/api/models/trial.js
@@ -34,6 +34,7 @@ const relatedModels = [
   'risks_of_bias',
   'risks_of_bias.source',
   'risks_of_bias.risk_of_bias_criteria',
+  'source',
 ];
 
 const Trial = BaseModel.extend({
@@ -59,6 +60,8 @@ const Trial = BaseModel.extend({
       Object.getPrototypeOf(Trial.prototype).serialize.call(this, args)
     );
     const relations = this.relations;
+    const ignoredRelations = ['source'];
+    const serializedRelations = _.difference(Object.keys(relations), ignoredRelations);
 
     attributes.locations = [];
     attributes.interventions = [];
@@ -67,7 +70,7 @@ const Trial = BaseModel.extend({
     attributes.organisations = [];
     attributes.risks_of_bias = [];
 
-    for (const relationName of Object.keys(relations)) {
+    for (const relationName of serializedRelations) {
       attributes[relationName] = relations[relationName].map((model) => {
         const attrs = model.toJSON();
 
@@ -121,6 +124,9 @@ const Trial = BaseModel.extend({
   risks_of_bias() {
     return this.hasMany('RiskOfBias');
   },
+  source() {
+    return this.belongsTo('Source', 'source_id');
+  },
   toJSONSummary() {
     const attributes = this.toJSON();
 
@@ -147,8 +153,12 @@ const Trial = BaseModel.extend({
       const robSources = this.related('risks_of_bias')
                              .toJSON()
                              .map((rob) => rob.source);
+      const trialSource = this.related('source').toJSON().id !== undefined ?
+            this.related('source').toJSON() :
+            undefined;
 
       const sources = [
+        trialSource,
         ...publicationsSources,
         ...documentsSources,
         ...recordsSources,
@@ -212,10 +222,12 @@ const Trial = BaseModel.extend({
       return discrepancies;
     },
     is_registered() {
-      // Be aware that this depends on `source_id` to be accurate
-      // So if a trial has a `nct` source but a `pubmed` source_id, the bug is elsewhere
-      const source = this.sources[this.attributes.source_id];
-      return source.type === 'register';
+      let status;
+      const source = this.sources[this.attributes.source_id] || { type: null };
+      if (source.type !== null) {
+        status = source.type === 'register';
+      }
+      return status;
     },
   },
 }, {
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -519,6 +519,9 @@ definitions:
       has_published_results:
         type: boolean
         description: Trial has its results published (true/false)
+      is_registered:
+        type: boolean
+        description: Whether the trial is registered in a clinical trial registry
       registration_date:
         type: string
         description: Date the trial was registered
--- a/test/api/models/trial.js
+++ b/test/api/models/trial.js
@@ -32,8 +32,15 @@ describe('Trial', () => {
       'risks_of_bias',
       'risks_of_bias.source',
       'risks_of_bias.risk_of_bias_criteria',
+      'source',
     ]);
   });
+
+  it('does not return its own source directly', () => (
+    factory.create('source', { id: 'nct' })
+      .then(() => factory.create('trial', { source_id: 'nct' }))
+      .then((trial) => (should(trial.toJSON().source).equal(undefined)))
+  ));
 
   describe('locations', () => {
     it('is an empty array if there\'re none', () => {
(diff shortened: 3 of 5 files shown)


api/models/trial.js
```diff
@@ -210,6 +211,13 @@ const Trial = BaseModel.extend({
 
       return discrepancies;
     },
+    is_registered() {
+      // Be aware that this depends on `source_id` to be accurate
+      // So if a trial has a `nct` source but a `pubmed` source_id, the bug is elsewhere
+      const registers = _.filter(this.sources, (source) => source.type === 'register')
+            .map((source) => source.id);
+      return registers.indexOf(this.attributes.source_id) > -1;
```
vitorbaptista: Wouldn't it be easier to do something like:

```javascript
const source = this.sources[this.attributes.source_id];
return source.type === 'register';
```

api/swagger/swagger.yaml
```diff
@@ -519,6 +519,9 @@ definitions:
       has_published_results:
         type: boolean
         description: Trial has its results published (true/false)
+      is_registered:
+        type: boolean
+        description: Whether the source is a non registry or not
```
vitorbaptista: The description feels too focused on the implementation details. Something like `Whether the trial is registered in a clinical trial registry` is clearer IMHO.

api/models/trial.js
nightsh: I think at the point when I was writing this, I didn't have `type` exposed yet. Sounds like a sensible approach, trying it out now ;)

api/swagger/swagger.yaml
nightsh: Cool, I'm so copying this into swagger 😀

api/models/trial.js
nightsh: I totally forgot about these cute little shorthands, kudos for reminding me about them! Pushing the new code in a minute.

vitorbaptista:
@nightsh I rebased master into this branch and found a bug on `trial.sources`. I added a test to reproduce at https://github.com/opentrials/api/pull/140/commits/94ad8aaf892e4dde078991d790c6e23d1fbdd599, please take a look.

This bug is unrelated to your code, and will be tricky to fix as we're overloading `trial.sources` with a virtual attribute. If it takes too long, I'm happy to move on with this current PR (after fixing the other issues) and opening a separate issue to deal with it.

nightsh:
I tried to return a trial's own `source_id` in the `sources` virtual, if no source has that ID already. Here are all the bumps I've hit while doing it:

* **Testing is terrible.** We currently have uuids as `source.id` attributes. This leads to a cascade of problems while trying to accurately simulate real world sources, amongst the most notable is that you can't instantiate a trial with a given `source_id` without violating the FK constraint. I tried to fix it quickly, but none of my hacks were enough. We need to change the factory to reflect reality. I had blockers around this, so I got stuck and halted.

* **Simple fixes are not enough.** I ran into long loops and weird conditions nesting while trying to gracefully overload the already overloaded `sources`.

* **I ran into specs ambiguity.** I tried to append a source stub to the virtual. Because if we can't rely on the `source_id` to be amongst `sources` virtual, then we can't rely on it to be any existing source and fetch it. Or can we? So the stub is to be added only if the `source` with that ID does not exist. It gets complex on ambiguous grounds, so I decided to stop until we discuss it further.

Have any unblocking fix for it?

Another issue is that the counterpart of this PR, i.e. the explorer one, is already merged (I probably got trigger happy after its approval). So explorer deployment is now blocked by this PR, sorry... 😞 The point is we need to move it somehow.

vitorbaptista:
@nightsh So I think the least bad way to move forward with this is:

```diff
diff --git a/api/models/trial.js b/api/models/trial.js
index 5a930ec..ee9ffbf 100644
--- a/api/models/trial.js
+++ b/api/models/trial.js
@@ -34,6 +34,7 @@ const relatedModels = [
   'risks_of_bias',
   'risks_of_bias.source',
   'risks_of_bias.risk_of_bias_criteria',
+  'source',
 ];
 
 const Trial = BaseModel.extend({
@@ -68,6 +69,9 @@ const Trial = BaseModel.extend({
     attributes.risks_of_bias = [];
 
     for (const relationName of Object.keys(relations)) {
+      if (relationName == 'source') {
+        continue;
+      };
       attributes[relationName] = relations[relationName].map((model) => {
         const attrs = model.toJSON();
 
@@ -90,6 +94,9 @@ const Trial = BaseModel.extend({
 
     return attributes;
   },
+  source() {
+    return this.belongsTo('Source', 'source_id');
+  },
   locations() {
     return this.belongsToMany('Location', 'trials_locations',
       'trial_id', 'location_id').withPivot(['role']);
@@ -149,6 +156,7 @@ const Trial = BaseModel.extend({
                              .map((rob) => rob.source);
 
       const sources = [
+        this.related('source').toJSON(),
         ...publicationsSources,
         ...documentsSources,
         ...recordsSources,
```

This ends up also adding `trial.source` to the results, which should be removed by removing it from the `visible` array. This would require a small refactoring, as we're adding all `relatedModels` there. The `if (relationName == 'source')` is quite messy as well. Actually, the entire `serialize()` method is messy. We only need it there to add empty arrays by default even when a trial doesn't have a relationship to an entity, so clients can always do `trial.conditions.map()` without having to test if `trial.conditions` exist in the first place.

I think that's a small benefit for the mess it creates, and I created an issue some time ago to modify it (https://github.com/opentrials/opentrials/issues/538). We can leave that in for now, possibly instead of explicitly checking that `relationName == 'source'`, we can check if `relations[relationName]` is an array.

And, as with anything, tests!

What do you think?

test/api/models/trial.js
```diff
@@ -300,6 +301,13 @@ describe('Trial', () => {
           .then((trialId) => new Trial({ id: trialId }).fetch({ withRelated: ['documents', 'documents.source'] }))
           .then((trial) => should(trial.toJSON().sources).deepEqual({}))
       ));
+
+      it('returns its own source', () => (
+        factory.create('source', { id: 'nct' })
+          .then(() => factory.create('trialWithRelated', { source_id: 'nct' }))
+          .then((trial) => (should(trial.toJSON().sources).have.key(trial.attributes.source_id))
+          )
+      ));
```
vitorbaptista: Instead of creating a `trialWithRelated` here, use a regular `trial`, otherwise you can't be sure if the source comes from the trial itself or some of its related entities.

vitorbaptista: It's also missing a few tests. For instance, what happens if you do:

```javascript
factory.create('trial')
  .then((trial) => new Trial({ id: trial.attributes.id }).fetch())
  .then((trial) => should(trial.toJSON()).be.undefined())
```

As in the trial's source wasn't loaded. Would it raise an exception? It's a bit tricky to know what `trial.is_registered` should be in this case. I considered `undefined`, as we can't say if it's registered or not because we don't have information on its source.

vitorbaptista: You're also missing a test that `trial.source` is undefined even if it has a source and it was loaded.

nightsh: On `trialWithRelated` vs `trial`, since the `source_id` I'm using is nowhere in the factory, there's no real danger. But as far as testing goes, I consider both variants equal. I will go back to simple `trial` though, there's no need for relations (anymore, it was at some point in my adventures).

On unloaded `trial.source`, I agree with `undefined` value. Makes sense.

I didn't consider testing an eventual `source` fail, thanks for catching that. It will not escape this time :)

api/models/trial.js
```diff
@@ -210,6 +221,17 @@ const Trial = BaseModel.extend({
...
+    is_registered() {
+      // Be aware that this depends on `source_id` to be accurate
+      // So if a trial has a `nct` source but a `pubmed` source_id,
+      // then the problem is with the processor
+      let status;
+      const source = this.sources[this.attributes.source_id] || { type: null };
+      if (source.type !== null) {
+        status = source.type === 'register';
+      }
+      return status;
```
vitorbaptista: I think it would be better to remove this comment. Here we assume all data is correct, otherwise we're displaying incorrect information.

This code could also be simplified to:

```javascript
const source = this.sources[this.attributes.source_id] || {};
return (source.type === 'register');
```

Without loss of readability IMHO. Also, in general, it's clearer to use parenthesis when writing statements like `a = b == c`, as it's a tricky one (i.e. is it `(a = b) == c` or `a = (b == c)`?)

nightsh: Good points, except that node would complain when no source is returned: trying to access `source.type` when `source` is `undefined` (that's why I place a stub `source` having only `type` if none found. I cannot reproduce this error in the node console though, so I guess it has something to do with one of the many libraries in a request's life cycle.

I'm doing the others now, but I don't guarantee I will be successful removing the `{ type: null }` object from there. I'll try again though, see what happens (maybe the error was somewhere else).

nightsh: Also that comparison does not return `undefined` when there's no source, which is what we want it to do, according to a discussion on earlier commits here. So, when a boolean is not enough, we use local vars :smile:

PR #141, opened 2017-01-30
[#531] Upgrade to Node 6.9.1

nightsh:
Fixes opentrials/opentrials#531


nightsh:
@vitorbaptista can we merge this?

PR #143, opened 2017-01-30
[#650] Replace document type with document category

georgiana-b:
opentrials/opentrials#650

### DIFF
--- a/migrations/20170125115201_create_document_categories.js
+++ b/migrations/20170125115201_create_document_categories.js
@@ -1,0 +1,173 @@
+'use strict';
+
+const documentCategories = [
+  {
+    id: 19,
+    name: 'Registry entry',
+    group: null,
+  },
+  {
+    id: 20,
+    name: 'Other',
+    group: null,
+  },
+  {
+    id: 21,
+    name: 'Journal article',
+    group: 'Results',
+  },
+  {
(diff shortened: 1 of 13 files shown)


migrations/20170125115201_create_document_categories.js
```diff
@@ -0,0 +1,173 @@
+'use strict';
+
+const documentCategories = [
+  {
+    id: 19,
```
nightsh: I really don't like the idea of hardcoding IDs into migrations. Isn't there any other way of doing it? Maybe we can further match explorer and API categories by name+group pairs, which are unique.

migrations/20170126145822_replace_document_type_with_document_categories.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+const _ = require('lodash');
+const Promise = require('bluebird');
+
+const typeToCategoryMapping = {
+  blank_consent_form: 33,
```
nightsh: Same issue I have with hardcoded IDs into migrations.

georgiana-b: @nightsh I figured the `name` and `group` would probably change more often than the ID, wouldn't you agree?

PR #145, opened 2017-02-15
[#650] Fix migration that replaces document type for category

vitorbaptista:
We can't add a non-nullable column without a default value to an existing table.
To do so, we need to add a nullable column, then set the data as we'd like, and
after that sets it as non-nullable.

opentrials/opentrials#650


vitorbaptista:
@georgiana-b please review

PR #146, opened 2017-02-21
Add `document_categories` endpoint

nightsh:
As described in opentrials/opentrials#685, we now have an endpoint for exposing
the document categories from the API instead of storing them in Explorer.


nightsh:
I have to add some tests and it'll be ready for review.

PR #147, opened 2017-02-23
[WIP] [#676] Create audit log for trial deduplication process

arthurSena:
As described in [#676](https://github.com/opentrials/opentrials/issues/676), we built a new table in the database in order to store the information about how the trial deduplication was made. The table follows the format below: 

trial_id | record_id | method | commit | created_at | updated_at
------------ | -------------|------------ | -------------|------------ | -------------
A | B|identifier | COMMIT_HASH
A | C|title |  COMMIT_HASH


migrations/20170216104317_create_audit_log_for_trial_deduplication.js
```diff
@@ -0,0 +1,30 @@
...
+
+      table.timestamps(true, true);
+
+    })
+    .raw(`ALTER TABLE trial_deduplication_logs ADD CONSTRAINT   
+            trial_deduplication_logs_trials_id_foreign  FOREIGN KEY 
+            (trial_id) REFERENCES trials (id);`)
+    .raw(`ALTER TABLE trial_deduplication_logs ADD CONSTRAINT  
+            trial_deduplication_logs_records_id_foreign FOREIGN KEY 
+            (record_id) REFERENCES records (id);`)
```
vitorbaptista: No need to add the foreign key constraints manually here, as Knex has a helper method for this. Check https://github.com/opentrials/api/blob/master/migrations/20160216120159_create_initial_schema.js#L68-L70

migrations/20170216104317_create_audit_log_for_trial_deduplication.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema
+    .createTable('trial_deduplication_logs', (table) => {
+      table.uuid('record_id')
+        .primary();
```
vitorbaptista: `record_id` can't be the primary key, otherwise it won't be possible to implement the pattern described on https://github.com/opentrials/opentrials/issues/676#issuecomment-280060487. More specifically, allowing us to know not only what was the latest match a record had, but the previous ones as well. For example:

trial_id | record_id | method | commit | created_at | updated_at
------------ | -------------|------------ | -------------|------------ | -------------
A | B|title | | 2017-02-01 | |
A | B|identifier |  | 2017-02-05 | |

Maybe not, but I'd imagine it would require a separate `id` field (can be a simple auto-incrementing int).

arthurSena: Yes, you're right. Fixing now...

vitorbaptista:
All good, just need to fix linting.

vitorbaptista:
Will be fixed on https://github.com/opentrials/api/pull/153

PR #148, opened 2017-02-23
[#693] Fix issue when reindexing trials and stop indexing on error

vitorbaptista:
Fixes opentrials/opentrials#693


vitorbaptista:
Merging this without review as production is currently offline because of it.

PR #151, opened 2017-03-22
[#709] Expose phase data

nightsh:
Fixes opentrials/opentrials#709

### DIFF
--- a/api/swagger/swagger.yaml
+++ b/api/swagger/swagger.yaml
@@ -551,6 +551,11 @@ definitions:
         type: string
         description: Date when a certification of exemption has been filed for the trial
         format: date-time
+      study_phase:
+        type: array
+        description: Phases of the study (e.g. `["Phase 2"]` or `["Phase 1", "Phase 2"]`)
+        items:
+          type: string
       status:
         type: string
         description: Completion status of the trial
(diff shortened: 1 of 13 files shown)


api/swagger/swagger.yaml
```diff
@@ -551,6 +551,11 @@ definitions:
         type: string
         description: Date when a certification of exemption has been filed for the trial
         format: date-time
+      study_phase:
+        type: array
+        description: Phases of the study
+        items:
+          type: string
```
vitorbaptista: Would be useful to add an example on how it looks like as well.

nightsh: Yeah, that's because I wasn't sure at the moment, with all the variations :sweat_smile:
I'll add 1-2 samples now, that I've read the code for the phase normalisation PR :+1:

nightsh:
@vitorbaptista we're currently lacking a test for checking if all the models' visible properties are added to the payload. I would make the needed changes for this to happen, but feels a bit like solving a separate issue.
Let me know what you think re:testing, should we test if all the models' attributes are added to the final response, for all models, explicitly?

nightsh:
@vitorbaptista ready for another review!

PR #152, opened 2017-03-30
[#646] Stop reindex at ElasticSearch error and keep old index

georgiana-b:
Fixes opentrials/opentrials#646

Desired changes:
- stop `reindex` if indexing errors are encountered
- delete the new (unfinished) index and fallback to the old one if errors are encountered
- show errors

To test add `dynamic: 'strict',` [in the `trials` index mapping](https://github.com/opentrials/api/blob/master/tools/indexers/trials.js#L17) (on the same level with `properties`).

### DIFF
--- a/tools/indexers/helpers.js
+++ b/tools/indexers/helpers.js
@@ -2,6 +2,7 @@
 
 'use strict';
 
+const _ = require('lodash');
 const Promise = require('bluebird');
 const client = require('../../config').elasticsearch;
 
@@ -39,6 +40,11 @@ function indexModel(model, index, indexType, _queryParams, fetchOptions, entitie
         .then((entities) => _bulkIndexEntities(entities, index, indexType))
         // eslint-disable-next-line no-loop-func
         .then((resp) => {
+          if (resp && resp.errors) {
+            const failedModels = _.filter(resp.items, 'index.error');
+            console.error(`${failedModels.length} failed to reindex.`);
+            throw _.map(failedModels, (item) => item.index);
+          }
           const count = (resp) ? resp.items.length : 0;
           numReindexedModels += count;
           console.info(`${numReindexedModels} successfully reindexed, ${modelCount - numReindexedModels} remaining.`);
@@ -47,11 +53,7 @@ function indexModel(model, index, indexType, _queryParams, fetchOptions, entitie
       offset += batchSize;
     } while (offset <= modelCount);
 
-    return chain
-      .catch((err) => {
-        console.error(err);
-        process.exit(-1);
-      });
+    return chain;
   });
 }
 
--- a/tools/reindex.js
+++ b/tools/reindex.js
@@ -19,13 +19,12 @@ function runIndexer(indexDefinition, alias, indexer) {
   return Promise.resolve()
     .then(() => client.indices.create(mapping))
     .then(() => indexer(index))
+    .catch((err) => removeIndexes([index]).then(() => { throw err; }))
+    .then(() => updateAlias(index, alias))
     .catch((err) => {
-      const errors = [].concat(err);
-      errors.map((error) => console.error(error));
-      removeIndexes([index])
-        .then(() => process.exit(-1));
-    })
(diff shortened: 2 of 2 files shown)


tools/indexers/helpers.js
```diff
@@ -39,6 +40,11 @@ function indexModel(model, index, indexType, _queryParams, fetchOptions, entitie
         .then((entities) => _bulkIndexEntities(entities, index, indexType))
         // eslint-disable-next-line no-loop-func
         .then((resp) => {
+          if (resp && resp.errors) {
+            const failedModels = _.filter(resp.items, 'index.error');
+            console.error(`${failedModels.length} failed to reindex.`);
+            throw _.map(failedModels, (item) => item.index);
+          }
```
vitorbaptista: According to the docs, `client.bulk()`, which is what's being returned by `_bulkIndexEntities()`, returns a promise that is rejected on errors. So, if an error occurred, the `.then()` block would never be executed, so this `if (resp && resp.errors)` block does nothing.

What am I missing?

tools/reindex.js
```diff
@@ -19,6 +19,12 @@ function runIndexer(indexDefinition, alias, indexer) {
   return Promise.resolve()
     .then(() => client.indices.create(mapping))
     .then(() => indexer(index))
+    .catch((err) => {
+      const errors = [].concat(err);
+      errors.map((error) => console.error(error));
+      removeIndexes([index])
+        .then(() => process.exit(-1));
+    })
```
vitorbaptista: The `.catch()` should be the last thing we do in the Promise chain, so it catches all errors. We can't simply move this `.catch()` as is, though, as if the error happened during `updateAlias()`, we don't want to remove the indexes. This gives us two options: don't remove the indexes (they'll be removed on the next reindex run), or use two `.catch()`. We could change it to something like:

```python
    # ...
    .then(() => indexer(index))
    .catch((err) => removeIndexes([index]).then(() => { throw err }))
    .then(() => updateAlias(index, alias))
    .catch((err) => {
      console.error(err);
      process.exit(-1);
    });
```

This way, if the error was on the `indexer()` call or before, we try to remove the newly created index and rethrow the original error. It'll go down the chain, skipping all `.then()` (i.e. the `updateAlias()`) and end up in the last `.catch()`, where we print the error and exit. If the error is in `updateAlias()`, we just print the error and exist (i.e. don't remove the index).

Note that we're returning the `removeIndexes()`. This is necessary because `removeIndexes()` returns a promise, which needs to be part of the chain to be guaranteed to run.

tools/indexers/helpers.js
georgiana-b: `client.bulk()` returns the following `Object` on error:

```json
{
  "took": 12,
  "errors": true,
  "items": [
    {
      "index": {
        "_index": "trials_2017-04-05_1491385001035",
        "_type": "trial",
        "_id": "05cc77ad-5575-4c04-9309-4c64d5144b07",
        "status": 400,
        "error": {
          "type": "strict_dynamic_mapping_exception",
          "reason": "mapping set to strict, dynamic introduction of [identifiers] within [trial] is not allowed"
        }
      }
    },
    {
      "index": {
        "_index": "trials_2017-04-05_1491385001035",
        "_type": "trial",
        "_id": "475456f3-23bc-4f5e-9d19-51f4a1165540",
        "status": 400,
        "error": {
          "type": "strict_dynamic_mapping_exception",
          "reason": "mapping set to strict, dynamic introduction of [identifiers] within [trial] is not allowed"
        }
      }
    }
  ]
}
```

The docs for the latest version [include an example response of the Bulk API](https://www.elastic.co/guide/en/elasticsearch/reference/5.x/docs-bulk.html) which looks the same as :point_up:. The [docs for our version](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-bulk.html) don't include example response but they do mention: 

> The response to a bulk action is a large JSON structure with the individual results of each action that was performed. The failure of a single action does not affect the remaining actions.

PR #155, opened 2017-04-06
[#725] Add migration to set study_phase = NULL where it's [null]

vitorbaptista:
The migration that converted `study_phase` to be a `text[]` converted the
current study phases to a single-element array, including null study phases.
Because of this, we ended up with invalid study phases like `[null]`.

Fixes opentrials/opentrials#725


vitorbaptista:
This is currently breaking production, so I'm merging it. @nightsh please take a look when you're free.

nightsh:
Ah, I thought you merged it already. Sure, no problem :v:

PR #156, opened 2017-04-07
[#647] Strict mapping for ElasticSearch

georgiana-b:
Fixes opentrials/opentrials#647

### DIFF
--- a/tools/indexers/trials.js
+++ b/tools/indexers/trials.js
@@ -3,6 +3,31 @@
 const esHelpers = require('./helpers');
 const Trial = require('../../api/models/trial');
 
+const sourceMapping = {
+  properties: {
+    id: {
+      type: 'string',
+      index: 'not_analyzed',
+    },
+    name: {
+      type: 'string',
+      index: 'not_analyzed',
+    },
+    source_url: {
+      type: 'string',
+      index: 'not_analyzed',
+    },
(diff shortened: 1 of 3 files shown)


tools/indexers/trials.js
```diff
@@ -332,11 +332,16 @@ const trialMapping = {
     },
     discrepancies: {
       properties: {
-        public_title: getDiscrepancyRecordMapping({
```
georgiana-b: @vitorbaptista I updated these attributes to bring them in line with [`discrepancyFields` in `trial` model](https://github.com/opentrials/api/blob/master/api/models/trial.js#L187). I already run `reindex` with this mapping on a production dump and it didn't raise errors which means these fields (`public_title` etc.) are not used anymore. Can you remember any reason for them?

georgiana-b: Found [my answer](https://github.com/opentrials/api/commit/cc964c3b7180460ff4499dbd75cad823309b033d#diff-7b1832e030ab3a88fe949f72ada33c0bL114) and it appears the index attributes were missed when the model attributes were updated. So, this is safe to merge.

vitorbaptista: Agreed. Looks like I missed the mappings here when changed the discrepancy fields :+1:

PR #159, opened 2017-04-11
[#708] Change PubMed source type to 'journal'

nightsh:
This will only apply to PubMed at the moment.

Related to opentrials/opentrials#708, in which we need to find items coming from a register or a journal publication.


vitorbaptista:
Already merged on https://github.com/opentrials/api/pull/157

PR #161, opened 2017-04-20
[#757] Record's `is_primary` should not be null

nightsh:
Linked to opentrials/opentrials#757

### DIFF
--- a/migrations/20170419122022_make-is_primary-not-nullable.js
+++ b/migrations/20170419122022_make-is_primary-not-nullable.js
@@ -1,0 +1,11 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema
+    .raw('UPDATE records SET is_primary = false where is_primary IS NULL')
+    .raw('ALTER TABLE records ALTER COLUMN is_primary SET NOT NULL')
+);
+
+exports.down = () => {
+  throw Error('Destructive migration can\'t be rolled back.');
+};


migrations/20170419122022_make-is_primary-not-nullable.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema
+    .raw('UPDATE records SET is_primary = false where is_primary IS NULL')
+    .raw('ALTER TABLE records ALTER COLUMN is_primary SET NOT NULL')
```
nightsh: Went for a raw query on altering the table, mostly because the `knex` variant seemed way more complex and harder to read. At least coming after the value alteration above.

migrations/20170419122022_make-is_primary-not-nullable.js
```diff
@@ -0,0 +1,12 @@
...
+);
+
+exports.down = (knex) => (
+  knex.schema
+    .raw('ALTER TABLE records ALTER COLUMN is_primary DROP NOT NULL')
```
nightsh: Didn't throw an error due to the destructive change in `up`, because the `ALTER TABLE` is reversible. Didn't look like a good candidate for splitting the migration in two different files though.

vitorbaptista: Although I understand why this can be useful, I don't think it's a good idea. The invariant of a down migration is that it must leave the database in the same state as it was before. We can't do it in this case, so I'd prefer to raise an Exception.

nightsh: Alright, will push an update now.

